PR #8, opened 2025-11-01
hipFile: Remove unused test files

derobins:
These files have been compile-only for some time.


kurtmcmillan (approved):

riley-dixon (approved):
I still use these programs for testing Nvidia details, but can accept if we want to remove them since they are not more widely used.

jordan-turbofish (commented):
Should we just make them example programs, since we will need examples?

derobins:
Closing since these are still in use

PR #12, opened 2025-11-03
rocfile: Move file inspection in DriverState::registerFile(...) into UnregisteredFile

kurtmcmillan:


rocfile/src/file.h
```diff
@@ -38,6 +38,42 @@ struct FileOperationsOutstanding : public std::runtime_error {
     }
 };
 
+class UnregisteredFile {
```
derobins, line 41: Do we want to try to maintain 1 class per file?

kurtmcmillan: Good question. We already have multiple classes in file.cpp and buffer.cpp (FileMap, and BufferMap).

My rational for putting UnregisteredFile in file.cpp is that I'm (poorly) borrowing the typestate pattern from Rust. A file starts as a UnregisteredFile and becomes a (Registered)File when it is successfully registered. In other words, a UnregisteredFile is a File but with its "unregistered" status/flag indicated by its type.

kurtmcmillan:
This PR is from my private fork of hipFile. Unfortunately, our CI is not compatible with this workflow. Moving this PR to #14 which is a branch with this repository.

rocfile/src/file.h
kurtmcmillan: Moved comment [here](https://github.com/ROCm/hipFile/pull/14#discussion_r2491198668) in #14

PR #14, opened 2025-11-04
rocfile: Move file inspection in DriverState::registerFile(...) into UnregisteredFile

kurtmcmillan:


rocfile/src/file.cpp
```diff
@@ -18,14 +18,45 @@ using std::vector;
 
 namespace rocFile {
 
+UnregisteredFile::UnregisteredFile(int fd)
```
kurtmcmillan, line 23: @derobins Moving your [comment](https://github.com/ROCm/hipFile/pull/12#discussion_r2488212018) from #12 here.

> Do we want to try to maintain 1 class per file?

Good question. We already have multiple classes in file.cpp and buffer.cpp (FileMap, and BufferMap).

My rational for putting UnregisteredFile in file.cpp is that I'm (poorly) borrowing the typestate pattern from Rust. A file starts as a UnregisteredFile and becomes a (Registered)File when it is successfully registered. In other words, a UnregisteredFile is a File but with its "unregistered" status/flag indicated by its type.

jordan-turbofish (approved):

derobins (approved):

PR #29, opened 2025-11-06
Add an include-what-you-use option

derobins:
Name: AIS_USE_IWYU (default OFF)


kurtmcmillan (approved):
Could we add this to CI?

derobins:
> Could we add this to CI?

I would not at this time. Let's see how it does with the suggestions first :D

It'd also be helpful to make everything IWYU-clean before turning it on as a CI check (we are nowhere close).

PR #30, opened 2025-11-07
Fix include-what-you-use issues

derobins:
IWYU is generally quiet after this patch, aside from:

* Google Test headers being declared local
* One instance of mountinfo.h being declared a system header
* IWYU wants to strip stdbool.h from hipfile.h and rocfile.h since it's running a C++ compiler

We can try to handle this later in a mapping file.


derobins:
Fun fact: `hip/hip_runtime_api.h` MUST come before `hip/driver_types.h` because the latter is not self-contained.

derobins:
FWIW, IWYU has a lot of bad ideas about C vs C++ and system vs local headers. It's not ready for being a GitHub CI action and we should definitely not use the auto-fix tool.

kurtmcmillan (approved):
If IWYU is not part of CI should we add a contributions guide stating that contributors should run IWYU for each commit?

PR #32, opened 2025-11-08
Rename LICENSE to LICENSE.md

derobins:
This seems to be the standard in TheRock repos


derobins:
Jenkins is sad about LICENSE being renamed. I can fix that when this goes in.

kurtmcmillan (approved):

PR #33, opened 2025-11-09
Add a CMake option to run clang-tidy

derobins:


derobins:
FWIW, we are pretty clean. If we want to run this as an action, though, we'll have to turn off checks for enum ranges since that flags a bunch of problems in our tests. We can address the few issues in separate PRs.

kurtmcmillan (approved):

PR #36, opened 2025-11-09
Move install instructions to INSTALL.md

derobins:
As things get more complicated, the install instructions will become too complicated to fit in README.md. This also seems to be where TheRock components seems to put their install instructions.


derobins:
This is just a copy. I didn't update the instructions.

kurtmcmillan (approved):

PR #44, opened 2025-11-11
CI: Run different workflow if new image needs to be built.

riley-dixon:
We would like to enable a fork-based development process. Our current CI does not allow this due to us building and pushing a new "development" CI image for every PR. Pushing this image to GHCR is a privileged action and is not enabled by default from external PRs. Using build artifacts instead is another solution that does not require these privileges but loses certain performance qualities like caching. In addition, we found that certain machines could take 30-60 minutes to download the CI image via build artifacts.

This approach uses GHCR but limits when we choose to build and push a new development CI image. This will only be triggered if the Dockerfile has been changed. The rest of the CI process will then use this new development image when building and testing. Since pushing to GHCR is a privileged action, any changes to the Dockerfile must be submitted internally within the repo.

If the Dockerfile's have not been changed, no image will be built and will use the latest CI image that is kept in sync with the develop branch.

Whenever a PR is accepted & merged into develop, a new 'latest' CI image will be built and published.


riley-dixon:
Will need to reconfigure the "Required" CI Checks if this PR is accepted.

kurtmcmillan (approved):

derobins (approved):

PR #62, opened 2025-11-18
rocfile: Use statx to get direct io alignment and size requirements

kurtmcmillan:


derobins:
Are the Jenkins runners missing a package?

```
../../../rocfile/src/backend/fastpath.cpp:132:36: error: use of undeclared identifier 'STATX_DIOALIGN'
```

kurtmcmillan:
> Are the Jenkins runners missing a package?
> 
> ```
> ../../../rocfile/src/backend/fastpath.cpp:132:36: error: use of undeclared identifier 'STATX_DIOALIGN'
> ```

No. STATX_DIOALIGN is only available since 6.1. Ubuntu 22, and possibly others, are still running with 5 series. I'll update the PR so that hipFile requires 4kib alignment if STATX_DIOALIGN isn't available.

derobins (approved):

PR #65, opened 2025-11-19
Fix cmake-lint problems

derobins:
* Add an rc file for running cmake-lint
    * Currently just suppresses line length warnings
* Fix identified issues
* Add suppression comments for FetchContent format issues


derobins:
The repo will be cmake-lint clean with these changes. When this goes in, I'll add a CI job to ensure it stays clean.

jordan-turbofish (approved):

PR #86, opened 2025-11-24
Add AIS_BUILD_INTEGER_SANITIZERS

derobins:
Adds an option to build integer sanitizers (clang-only)

### DIFF
--- a/cmake/AISSanitizers.cmake
+++ b/cmake/AISSanitizers.cmake
@@ -48,7 +48,7 @@ function(ais_add_sanitizers target)
             target_compile_options(${target} PRIVATE ${integer_sanitize_flags})
             target_link_options(${target} PRIVATE ${integer_sanitize_flags})
         else()
-            message(WARNING "The integer sanitizer is only available on clang")
+            message(FATAL_ERROR "The integer sanitizer is only available on clang")
         endif()
     endif()
 
--- a/cmake/AISSanitizers.cmake
+++ b/cmake/AISSanitizers.cmake
@@ -38,8 +38,7 @@ function(ais_add_sanitizers target)
     # the process if undefined integer behaviour occurs (like
     # wrapping a signed integer).
     if(AIS_BUILD_INTEGER_SANITIZERS)
-        set(compiler_id "${CMAKE_CXX_COMPILER_ID}")
-        if(compiler_id STREQUAL "Clang")
+        if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
             set(integer_sanitize_flags
                 -fsanitize=integer
                 -fsanitize-minimal-runtime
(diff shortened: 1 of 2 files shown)


cmake/AISSanitizers.cmake
```diff
@@ -30,4 +31,25 @@ function(ais_add_sanitizers target)
...
+    # This has a very small runtime penalty (<1%) and will kill
+    # the process if undefined integer behaviour occurs (like
+    # wrapping a signed integer).
+    if(AIS_BUILD_INTEGER_SANITIZERS)
+        set(compiler_id "${CMAKE_CXX_COMPILER_ID}")
+        if(compiler_id STREQUAL "Clang")
+            set(integer_sanitize_flags
+                -fsanitize=integer
+                -fsanitize-minimal-runtime
+                -fno-sanitize-recover
```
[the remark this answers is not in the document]
derobins: Fixed

derobins:
I want to get this to pass, then switch it on all the time in debug builds.

cmake/AISSanitizers.cmake
```diff
@@ -30,4 +31,25 @@ function(ais_add_sanitizers target)
...
+        if(compiler_id STREQUAL "Clang")
+            set(integer_sanitize_flags
+                -fsanitize=integer
+                -fsanitize-minimal-runtime
+                -fno-sanitize-recover=integer
+            )
+            target_compile_options(${target} PRIVATE ${integer_sanitize_flags})
+            target_link_options(${target} PRIVATE ${integer_sanitize_flags})
+        else()
+            message(WARNING "The integer sanitizer is only available on clang")
```
riley-dixon: We use `FATAL_ERROR` if `AIS_BUILD_SANITIZERS` and `AIS_BUILD_THREAD_SANITIZERS` are both ON. Should we do the same here if we are using not-Clang? If no, feel free to auto-resolve.

cmake/AISSanitizers.cmake
```diff
@@ -30,4 +31,25 @@ function(ais_add_sanitizers target)
...
     endif()
+
+    # clang-only
+    #
+    # This has a very small runtime penalty (<1%) and will kill
+    # the process if undefined integer behaviour occurs (like
+    # wrapping a signed integer).
+    if(AIS_BUILD_INTEGER_SANITIZERS)
+        set(compiler_id "${CMAKE_CXX_COMPILER_ID}")
+        if(compiler_id STREQUAL "Clang")
```
riley-dixon: Do you know if `amdclang++` maps to `Clang`?

riley-dixon (approved):

kurtmcmillan (approved):

cmake/AISSanitizers.cmake
derobins: I'll switch it to FATAL_ERROR like the other one

cmake/AISSanitizers.cmake
derobins: Yes. That's what I usually build and test with.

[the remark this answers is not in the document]
derobins: We don't support Apple, so no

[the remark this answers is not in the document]
derobins: Fixed

riley-dixon: Okay cool, thanks for letting me know.

PR #88, opened 2025-11-24
rocfile: Untangle rocFileIo(...) and Fallback backend tests.

kurtmcmillan:


rocfile/test/rocfile.cpp
```diff
@@ -12,23 +12,42 @@
...
-#include "rocfile.h"
+#include "msys.h"
+#include "rocfile-private.h"
 #include "rocfile-test.h"
+#include "rocfile.h"
+#include "state.h"
+#include "sys.h"
 
+#include <array>
+#include <errno.h>
```
derobins: `#include <cerrno>`

kurtmcmillan: Fixed.

riley-dixon (approved):
Looks good. Only thing I could comment on is whether we should update the test names in _rocfile/test/fallback.cpp_ from snake_case to UpperCamelCase to align with GTest recommendations, but not really a huge deal.

jordan-turbofish (approved):

PR #92, opened 2025-11-25
Remove BUILD_HIPFILE/ROCFILE

derobins:
These CMake options will be meaningless as we consolidate hipFile and rocFile. We will now always build both libraries on AMD and only hipFile on NVIDIA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,21 +156,19 @@ include(AISIWYU)
 #-------------------
 include(AISSanitizers)
 
-#----------------------------------
-# rocfile (default OFF for NVIDIA)
-#----------------------------------
+#-------------------
+# Platform defaults
+#-------------------
 if(CMAKE_HIP_PLATFORM STREQUAL "nvidia")
-    set(BUILD_ROCFILE_DEFAULT OFF)
+    set(BUILD_ROCFILE OFF)
 else()
-    set(BUILD_ROCFILE_DEFAULT ON)
+    set(BUILD_ROCFILE ON)
 endif()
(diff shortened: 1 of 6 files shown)


cmake/AISInstall.cmake
derobins, line 1: This looks like big changes, but it's the diff algorithm being confused by removing the if() and shifting everything over to the left + a few minor tweaks.

docs/CMakeLists.txt
```diff
@@ -51,15 +51,7 @@ function(ais_add_docs_target AIS_DOC_PRETTY_NAME)
     add_dependencies(doc ${AIS_DOC_TARGET_NAME})
 endfunction()
 
-if(BUILD_HIPFILE)
-    ais_add_docs_target("hipFile")
-endif()
+ais_add_docs_target("hipFile")
 if(BUILD_ROCFILE)
     ais_add_docs_target("rocFile")
 endif()
```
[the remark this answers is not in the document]
derobins: No

kurtmcmillan (approved):

CMakeLists.txt
```diff
@@ -156,21 +156,19 @@ include(AISIWYU)
...
 
-#---------
-# hipfile
-#---------
-option(BUILD_HIPFILE "Build hipfile library" ON)
-cmake_dependent_option(BUILD_AISCP "Build aiscp example program" ON "BUILD_HIPFILE" OFF)
+#-----------------------
+# aiscp example program
+#-----------------------
+option(BUILD_AISCP "Build aiscp example program" ON)
```
derobins, line 171: When we have more examples, we should just change this to BUILD_EXAMPLES.

PR #95, opened 2025-11-26
Remove rocFile

derobins:
Completely removes all traces of rocFile


kurtmcmillan (approved):
Should probably wait for at least another reviewer.

derobins:
> Should probably wait for at least another reviewer.

I think @jordan-turbofish and @riley-dixon should also sign off. This is a big change.

hipfile/src/amd_detail/rocfile.cpp
```diff
@@ -69,31 +69,22 @@ catch (...) {
...
-    return {hipFileInternalError, hipSuccess};
-}
-catch (const FileNotRegistered &) {
-    return {hipFileHandleNotRegistered, hipSuccess};
+    return;
 }
 catch (...) {
-    return handle_exception();
+    return;
 }
```
riley-dixon: Why remove the error handling here and changing this to return void?

hipfile/src/amd_detail/rocfile.cpp
```diff
@@ -207,16 +200,32 @@ catch (...) {
...
 {
-    return hipFileIo(IoType::Read, fh, buffer_base, size, file_offset, buffer_offset, getCachedBackends());
+    auto result =
+        hipFileIo(IoType::Read, fh, buffer_base, size, file_offset, buffer_offset, getCachedBackends());
+
+    if (result == -hipFileDriverNotInitialized) {
+        // Match cuFile behaviour
+        errno  = EINVAL;
+        result = -1;
+    }
```
riley-dixon: Is this really what happens? This goes against their own documentation that return -1 means a POSIX error.

hipfile/src/amd_detail/rocfile.cpp
riley-dixon: Okay, I understand the reasoning now.

I'm going to (unsuccessfully) argue that we should keep our error handling _somewhere_, despite needing this particular target for compatibility reasons.

riley-dixon (approved):
I'm impressed that you were able to do this so quickly.

Batch and CI changes look good.

jordan-turbofish (approved):

hipfile/src/amd_detail/rocfile.cpp
riley-dixon: After talking with @jordan-turbofish this is in fact what happens.

PR #100, opened 2025-11-27
CI: Fix bug in files-changed.sh

riley-dixon:
Glob patterns should not be quoted.

### DIFF
--- a/util/files-changed.sh
+++ b/util/files-changed.sh
@@ -23,8 +23,9 @@ fi
 mapfile -t changed_files <<< "${diff_output}"
 
 for changed_file in "${changed_files[@]}"; do
-    if [[ "${changed_file}" == "${CHECK_PATTERN}" ]]; then
-         FILES_CHANGED=1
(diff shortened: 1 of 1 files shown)


derobins (approved):

util/files-changed.sh
```diff
@@ -23,8 +23,9 @@ fi
 mapfile -t changed_files <<< "${diff_output}"
 
 for changed_file in "${changed_files[@]}"; do
-    if [[ "${changed_file}" == "${CHECK_PATTERN}" ]]; then
-         FILES_CHANGED=1
+    if [[ "${changed_file}" == ${CHECK_PATTERN} ]]; then
+        FILES_CHANGED=1
+        break
```
derobins, line 30: shellcheck is sad about the change

riley-dixon: shellcheck is dumb because that is actually what I want to happen

riley-dixon:
Verified a new CI build is actually triggered if a Dockerfile change is made. Removed that commit from the final PR (due to this PR coming from an external fork which causes a build failure).

PR #105, opened 2025-11-27
AIROCFILE-35 - CI: Add hipFile IO test using aiscp

kurtmcmillan:
Test hipFile IO using aiscp to ensure IO works in some capacity. Future changes will allow testing with specific backends.

### DIFF
--- a/util/ci-aiscp-test.sh
+++ b/util/ci-aiscp-test.sh
@@ -14,13 +14,13 @@ set -e
 
 trap 'if [ -e "$TMPDIR" ]; then rm -fr "$TMPDIR"; fi' EXIT
 
-TESTDIR=$1
+WORKDIR=$1
 AISCP=$2
 
-echo "Test directory: $TESTDIR"
+echo "Test directory: $WORKDIR"
 echo "aiscp: $AISCP"
 
-TMPDIR=$(mktemp --directory --tmpdir="$TESTDIR")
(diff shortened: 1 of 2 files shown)


util/ci-aiscp-test.sh
```diff
@@ -0,0 +1,42 @@
...
+# Used by CI to test hipFile IO using aiscp
+
+# Usage:
+# ci-aiscp-test.sh WORKDIR AISCP
+#   TESTDIR: A directory on an AIS supported filesystem backed by an AIS supported device
+#   AISCP: Path to the aiscp binary
+
+set -e
+
+trap 'if [ -e "$TMPDIR" ]; then rm -fr "$TMPDIR"; fi' EXIT
```
riley-dixon, line 17: I would consider adding a cleanup step in the CI YAML with the `always()` condition. I do not know if this TRAP is guaranteed to execute in all cases (e.g. if GitHub aborts this job mid-step).

While doing so, I would just be mindful as well that the directory created here will be root owned.

riley-dixon (approved):

util/ci-aiscp-test.sh
kurtmcmillan: Thank you for the suggestions.

- I've updated the github workflow to control the lifetime of the directory used for IO tests
- IO tests are launched from within the IO test directory

riley-dixon: Looks good, thanks!

PR #109, opened 2025-11-28
AIROCFILE-35 - CI: Add hipFile fio IO test

kurtmcmillan:


derobins (approved):

derobins:
We should think about breaking up that uber-action at some point

riley-dixon:
> We should think about breaking up that uber-action at some point

Ideally yes. We should get packaging sorted out first because then we can remove the redundant compile steps (which is why the system test stage is kind of monolithic at the moment). That will help us split up the system test step more effectively.

.github/workflows/build-ais.yml
```diff
@@ -285,6 +292,55 @@ jobs:
...
+            -t \
+            -w /ais/fio/build \
+            ${AIS_CONTAINER_NAME} \
+            /bin/bash -c '
+              ROCM=/opt/rocm \
+              HIPFILE=/ais/hipFile \
+              HIPFILELIB=$HIPFILE/build/hipfile/src/amd_detail/ \
+              HIP_PLATFORM=amd \
+              CFLAGS="-I$ROCM/include -I$HIPFILE/hipfile/include" \
+              LDFLAGS="-L$ROCM/lib -L$HIPFILELIB -Wl,-rpath,$HIPFILELIB" \
```
riley-dixon: IMO: My style preference is wrapping the variable names in `{}` brackets. But won't block the PR over this. Though I may change it later if you don't ;)

riley-dixon (approved):

riley-dixon:
Also it seems the CI emits these warnings as build FIO:
```
In file included from ../engines/libhipfile.c:19:
/ais/hipFile/hipfile/include/hipfile.h:395:51: warning: 'struct sockaddr' declared inside parameter list will not be visible outside of this definition or declaration
  395 |     int (*getRDMADeviceList)(void *handle, struct sockaddr **hostaddrs);
      |                                                   ^~~~~~~~
/ais/hipFile/hipfile/include/hipfile.h:399:79: warning: 'struct sockaddr' declared inside parameter list will not be visible outside of this definition or declaration
  399 |     int (*getRDMADevicePriority)(void *handle, char *, size_t, hoff_t, struct sockaddr *hostaddr);
      |                                                                               ^~~~~~~~
```

I'll leave it to you to decide what you would like to do with this knowledge.

.github/workflows/build-ais.yml
kurtmcmillan: Wrapped.

kurtmcmillan:
> Also it seems the CI emits these warnings as build FIO:
> 
> ```
> In file included from ../engines/libhipfile.c:19:
> /ais/hipFile/hipfile/include/hipfile.h:395:51: warning: 'struct sockaddr' declared inside parameter list will not be visible outside of this definition or declaration
>   395 |     int (*getRDMADeviceList)(void *handle, struct sockaddr **hostaddrs);
>       |                                                   ^~~~~~~~
> /ais/hipFile/hipfile/include/hipfile.h:399:79: warning: 'struct sockaddr' declared inside parameter list will not be visible outside of this definition or declaration
>   399 |     int (*getRDMADevicePriority)(void *handle, char *, size_t, hoff_t, struct sockaddr *hostaddr);
>       |                                                                               ^~~~~~~~
> ```
> 
> I'll leave it to you to decide what you would like to do with this knowledge.

I added an include for sys/socket.h to hipfile.h. This should silence these warnings.

hipfile/include/hipfile.h
```diff
@@ -9,6 +9,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <sys/socket.h>
```
derobins, line 12: We should not be using POSIX headers unprotected in a platform-independent header file. This needs an ifdef guard of some sort, either via compile-time test or via a POSIX level macro.

derobins (changes requested):
Can't use `sys/socket.h` unprotected

.github/workflows/build-ais.yml
```diff
@@ -278,7 +291,56 @@ jobs:
...
+                --direct=1 \
+                --gpu_dev_ids=0 \
+                --filename=fio.dat \
+                --size=1G \
+                --bs=1M \
+                --verify=md5 \
+                --numjobs=1 \
+                --name=write \
+                --rw=write \
+                --do_verify=0 \
```
jordan-turbofish, line 338: Should we have `do_verify=1` on the writes as well?  This won't run the fio_libhipfile_pre_write verify path, right?

hipfile/include/hipfile.h
kurtmcmillan: https://github.com/ROCm/hipFile/pull/114 for guarding the include of sys/socket.h

.github/workflows/build-ais.yml
kurtmcmillan: Could do it on the write, but I think doing it on the read is good enough to know if the PR breaks IO.

PR #110, opened 2025-11-29
Remove Buffer and File class friends

derobins:
I don't recall the reason for making the ctors private, requiring friend classes to instantiate them. I'm personally not a fan of breaking encapsulation with friend classes.

### DIFF
--- a/hipfile/src/amd_detail/buffer.cpp
+++ b/hipfile/src/amd_detail/buffer.cpp
@@ -6,6 +6,7 @@
 #include "buffer.h"
 #include "context.h"
 #include "hip.h"
+#include "passkey.h"
 #include "sys.h"
 
 #include <algorithm>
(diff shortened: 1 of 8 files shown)


jordan-turbofish:
As they are now, File and Buffer objects are guaranteed to be registered.  This change would make that a convention, rather than enforced by the compiler.

kurtmcmillan:
> As they are now, File and Buffer objects are guaranteed to be registered. This change would make that a convention, rather than enforced by the compiler.

I was making this point when you made your comment. There is one exception, we temporarily create a Buffer for a hipFileRead/hipFileWrite if the buffer is not registered.

derobins:
Article about this pattern here:

https://arne-mertz.de/2016/10/passkey-idiom/

And here:

https://stackoverflow.com/questions/3217390/clean-c-granular-friend-equivalent-answer-attorney-client-idiom/3218920

riley-dixon (commented):
Looks good to me. But I think Kurt/Jordan should give a thumbs up for this to go in.

jordan-turbofish (approved):

derobins:
I'm going to take a sec to rename Key to PassKey to emphasize that it's not a key like in a KV store or anything.

jordan-turbofish:
Do you want to make the same change to stream too?

kurtmcmillan (approved):

jordan-turbofish (approved):

PR #111, opened 2025-12-01
Use default dtors for BufferMap and FileMap

derobins:
Making virtual calls in destructors is a bad idea in C++ and clang-tidy complains about this. There's really no reason to walk the maps to free elements, as the elements are all shared pointers and will be cleaned up automatically as the reference counts drop to zero.

This PR replaces the explicit destructors with default destructors.

### DIFF
--- a/hipfile/src/amd_detail/buffer.cpp
+++ b/hipfile/src/amd_detail/buffer.cpp
@@ -156,29 +156,4 @@ BufferMap::clear()
     from_ptr.clear();
 }
 
-BufferMap::~BufferMap()
-{
-    try {
-        // Create a list of registered buffers without causing the use_count to increase
-        vector<const void *> buffer_ptrs;
-        buffer_ptrs.reserve(from_ptr.size());
-        transform(from_ptr.begin(), from_ptr.end(), std::back_inserter(buffer_ptrs),
-                  [](const auto &pair) { return pair.first; });
-
-        for (const auto &ptr : buffer_ptrs) {
-            // NOLINTNEXTLINE(clang-analyzer-optin.cplusplus.VirtualCall)
-            deregisterBuffer(ptr);
-        }
-    }
-    catch (...) {
-#ifndef NDEBUG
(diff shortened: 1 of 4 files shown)


derobins:
This will finally make clang-tidy happy and we'll be able to switch it on for CI

jordan-turbofish:
The deregister functions also check whether the files / buffers are still in use.  To maintain this behavior, we could iterate over the maps and check the use counts, logging if any are unexpectedly in use.

derobins:
> The deregister functions also check whether the files / buffers are still in use. To maintain this behavior, we could iterate over the maps and check the use counts, logging if any are unexpectedly in use.

I think we need to clarify the expected lifecycle behavior of the map classes. I feel like at least some of the behavior we've enshrined here is due to our attempts to follow NVIDIA's behavior when messing with the use counts and closing the driver. If the existing behavior is necessary, it should be documented and checked for correctness in the tests.

jordan-turbofish:
> I think we need to clarify the expected lifecycle behavior of the map classes. I feel like at least some of the behavior we've enshrined here is due to our attempts to follow NVIDIA's behavior when messing with the use counts and closing the driver. If the existing behavior is necessary, it should be documented and checked for correctness in the tests.

This has no relation to the driver use count.  This is checking whether any files / buffers are still in use at destruction time, which shouldn't be the case if everything has been cleaned up.

derobins:
> > I think we need to clarify the expected lifecycle behavior of the map classes. I feel like at least some of the behavior we've enshrined here is due to our attempts to follow NVIDIA's behavior when messing with the use counts and closing the driver. If the existing behavior is necessary, it should be documented and checked for correctness in the tests.
> 
> This has no relation to the driver use count. This is checking whether any files / buffers are still in use at destruction time, which shouldn't be the case if everything has been cleaned up.

Okay, but the only thing we can do at that point is crash. This seems like it needs to be handled higher up in the library, not at destruction time.

derobins:
Closing. Will replace with an updated PR.

PR #112, opened 2025-12-01
CI: Checkout hipfile source to ${GITHUB_WORKSPACE}/ais

kurtmcmillan:
This allows multiple repositories to be checked out at the same time.


kurtmcmillan:
This work will be done as part of #109

PR #113, opened 2025-12-02
Remove virtual calls in dtors

### LINKED ISSUE #99: [Issue]: Fix clang-tidy virtual destructor issues
### Problem Description

Building with clang-tidy reports a couple of problems with making virtual calls in destructors. This is due to the way we are using mocks in the tests.

```
/.../hipfile/src/amd_detail/buffer.cpp:169:13: warning: Call to virtual method 'BufferMap::deregisterBuffer' during destruction bypasses virtual dispatch [clang-analyzer-optin.cplusplus.VirtualCall]
  169 |             deregisterBuffer(ptr);
      |             ^~~~~~~~~~~~~~~~~~~~~
/.../hipfile/src/amd_detail/buffer.cpp:169:13: note: Call to virtual method 'BufferMap::deregisterBuffer' during destruction bypasses virtual dispatch
  169 |             deregisterBuffer(ptr);
      |             ^~~~~~~~~~~~~~~~~~~~~
/.../hipfile/src/amd_detail/file.cpp:158:13: warning: Call to virtual method 'FileMap::deregisterFile' during destruction bypasses virtual dispatch [clang-analyzer-optin.cplusplus.VirtualCall]
  158 |             deregisterFile(fh);
      |             ^~~~~~~~~~~~~~~~~~
/.../hipfile/src/amd_detail/file.cpp:158:13: note: Call to virtual method 'FileMap::deregisterFile' during destruction bypasses virtual dispatch
  158 |             deregisterFile(fh);
      |             ^~~~~~~~~~~~~~~~~~
```

### Operating System

Any

### CPU

Any

### GPU

Any

### ROCm Version

Any

### ROCm Component

_No response_

### Steps to Reproduce

Build with clang-tidy (AIS_USE_CLANG_TIDY=ON)

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

_No response_

### Additional Information

_No response_

derobins:
clang-tidy complains about using virtual calls in the FileMap and BufferMap destructors. The virtual calls in question attempt to shut down the map objects, which is unnecessary, as the maps contain shared pointers that will be cleaned when they go out of scope.

The functionality that we retained is complaining about any in-use File or Buffer objects when the associated maps are shut down. Previously, this would result in exit() being called in debug builds and silently ignoring the problem in release builds. The new behavior is to create syslog messages in all builds (like our Stream class).

Fixes #99

### DIFF
--- a/hipfile/src/amd_detail/file.cpp
+++ b/hipfile/src/amd_detail/file.cpp
@@ -151,7 +151,7 @@ FileMap::~FileMap()
     // Complain in the logs if we're shutting down a FileMap
     // with files that are still in use.
     for (const auto &p : from_fh) {
-        if (p.second.use_count() > 1) {
+        if (p.second.use_count() > 2) {
             Context<Sys>::get()->syslog(LOG_CRIT, "FileMap state is being destructed with in-use files.");
             return;
         }
--- a/hipfile/src/amd_detail/file.cpp
+++ b/hipfile/src/amd_detail/file.cpp
@@ -150,11 +150,15 @@ FileMap::~FileMap()
 {
     // Complain in the logs if we're shutting down a FileMap
     // with files that are still in use.
+    int count = 0;
     for (const auto &p : from_fh) {
         if (p.second.use_count() > 2) {
-            Context<Sys>::get()->syslog(LOG_CRIT, "FileMap state is being destructed with in-use files.");
-            return;
+            count++;
         }
+    }
+
+    if (count > 0) {
+        Context<Sys>::get()->syslog(LOG_CRIT, "FileMap state is being destructed with in-use files");
     }
 }
 
--- a/hipfile/src/amd_detail/file.cpp
+++ b/hipfile/src/amd_detail/file.cpp
@@ -14,6 +14,7 @@
 #include <iterator>
 #include <utility>
 #include <sys/sysmacros.h>
+#include <syslog.h>
 #include <vector>
 
 using std::optional;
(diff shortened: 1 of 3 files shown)


derobins:
I'm not opposed to Copilot's suggestions. We should make all the classes uniform, though (i.e., changes here should also be made in Stream's dtor.

hipfile/src/amd_detail/file.cpp
```diff
@@ -147,27 +148,14 @@ FileMap::clear()
...
-        transform(from_fh.begin(), from_fh.end(), std::back_inserter(file_handles),
-                  [](const auto &pair) { return pair.first; });
-
-        for (const auto &fh : file_handles) {
-            // NOLINTNEXTLINE(clang-analyzer-optin.cplusplus.VirtualCall)
-            deregisterFile(fh);
+    // Complain in the logs if we're shutting down a FileMap
+    // with files that are still in use.
+    for (const auto &p : from_fh) {
+        if (p.second.use_count() > 1) {
```
jordan-turbofish: This needs to be `> 2` because each shared_ptr is in two maps.

derobins: Fixed

kurtmcmillan:
> I'm not opposed to Copilot's suggestions. We should make all the classes uniform, though (i.e., changes here should also be made in Stream's dtor.

It might be useful to log the file descriptor of each file still in use and the buffer pointer for each buffer still in use.

derobins:
> > I'm not opposed to Copilot's suggestions. We should make all the classes uniform, though (i.e., changes here should also be made in Stream's dtor.
> 
> It might be useful to log the file descriptor of each file still in use and the buffer pointer for each buffer still in use.

I thought about something like that, but how useful is that information?

kurtmcmillan:
> > > I'm not opposed to Copilot's suggestions. We should make all the classes uniform, though (i.e., changes here should also be made in Stream's dtor.
> > 
> > 
> > It might be useful to log the file descriptor of each file still in use and the buffer pointer for each buffer still in use.
> 
> I thought about something like that, but how useful is that information?

Good point. It may not be useful on its own. Perhaps when we have more logging. For now I guess it would be good enough just to indicate that there were outstanding operations on files/buffers.

kurtmcmillan (approved):

jordan-turbofish (approved):

riley-dixon (approved):

PR #117, opened 2025-12-05
Tidy hipfile.h

derobins:
* Add parameter struct member Doxygen
* Remove some unnecessary comments


hipfile/include/hipfile.h
```diff
@@ -360,10 +360,6 @@ typedef struct hipFileDriverProps {
 //  RDMA API
 // ***********************************************************************
 
-/*
- * NOTE: !! AIS UNSUPPORTED !! (Not related to NVMeoRDMA)
- */
```
derobins, line 365: This is probably not the best place to document this and we've removed all the other "not supported" comments

kurtmcmillan (approved):

PR #118, opened 2025-12-05
hipfile: Introduce Configuration

kurtmcmillan:
## Motivation

Introduce infrastructure that will allow hipFile behavior to be controlled with environment variables and settings in configuration files.

This patch series allows users to specify HIPFILE_ALLOW_COMPAT_MODE=false to force IO to take the AMD Infinity Storage (AIS) path path or, HIPFILE_FORCE_COMPAT_MODE=true to force IO to use the POSIX path.

### DIFF
--- a/hipfile/src/amd_detail/configuration.cpp
+++ b/hipfile/src/amd_detail/configuration.cpp
@@ -1,0 +1,37 @@
+/* Copyright (c) Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include "configuration.h"
+#include "environment.h"
+#include "hip.h"
+
+using namespace hipFile;
+
+Configuration::Configuration() : m_fastpath(true), m_fallback(true)
+{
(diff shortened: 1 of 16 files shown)


hipfile/src/amd_detail/configuration.cpp
```diff
@@ -0,0 +1,37 @@
...
+using namespace hipFile;
+
+Configuration::Configuration() : m_fastpath(true), m_fallback(true)
+{
+    auto maybe_env_force_compat{Environment::force_compat_mode()};
+    if (maybe_env_force_compat && maybe_env_force_compat.value()) {
+        m_fastpath = false;
+    }
+    m_fastpath = !!getHipAmdFileReadPtr() && m_fastpath;
+    m_fastpath = !!getHipAmdFileWritePtr() && m_fastpath;
```
kurtmcmillan, line 19: We do it this way so that `getHipAmdFileReadPtr` and `getHipAmdFileWritePtr` are always called. It simplifies test logic.

jordan-turbofish (approved):

PR #122, opened 2025-12-08
Add a packaging step

riley-dixon:
Add a step to create DEB & RPM packages of hipFile.

AIROCFILE-97

This currently only supports the default DEV release of hipFile. This will be a multi-PR endeavor to fully enable a flexible packaging process.


.github/workflows/build-ais.yml
```diff
@@ -105,6 +105,32 @@ jobs:
...
+            -t \
+            -w /ais/hipFile/build \
+            ${AIS_CONTAINER_NAME} \
+            /bin/bash -c '
+              cpack -G ${{ inputs.platform == 'ubuntu' && 'DEB' || 'RPM' }}
+            '
+      # CI can't really peer into CMake to capture this data ahead of time.
+      # We should only have a single package built so we can reasonably
+      # Just use a glob pattern on the expected package.
+      # For now, hardcode to the default DEV release.
```
kurtmcmillan: `built so we can reasonably Just use`

This is a little awkward.

Also, what do you mean 'hardcode to the default DEV release'?

.github/workflows/build-ais.yml
```diff
@@ -1,6 +1,7 @@
 name: build_and_test # Mono-workflow - May be advantageous to split up
 run-name: Build, Test, and Analyze AIS
 env:
+  AIS_GHA_CMD_FILES_DOCKER_DIR: /root/github
```
kurtmcmillan, line 4: General question about this commit.

Is this directory exclusively for this runner and workflow? Would other runners running on the same machine get their own directories? Are the contents of the directory wiped after each run?

kurtmcmillan (approved):

.github/workflows/build-ais.yml
riley-dixon: My understanding is that by default CMake will build hipFile with Development flags. So, the resulting package will be `hipfile-dev` or `hipfile-devel`.

Will fix the wording, thanks.

.github/workflows/build-ais.yml
riley-dixon: This path is inside the docker container, so it is not shared at all. If a self-hosted runner is trying to peer into a container it does not own, that is bad and the runner should not do that.

riley-dixon: > Are the contents of the directory wiped after each run?

To also address this question, yes. The container is destroyed at the end of the run and any changes made to the container are not preserved.

riley-dixon: In case you were talking about the actual directory on the host, this directory is contained within the runner's workspace. Similarly, another runner should not be peering into another runner's workspace.

The lifetime of this directory is on a per-workflow basis and is cleaned up at the end of the workflow.

kurtmcmillan: Thanks!

.github/workflows/build-ais.yml
kurtmcmillan: Thanks!

derobins (approved):

PR #125, opened 2025-12-10
hipfile: Use internal file descriptors for IO

kurtmcmillan:
## Motivation

No longer require files to have the O_DIRECT flag set when they are registered with hipFile. This change will also enable hipFile to eventually handle unaligned IO.

## Technical Details

When a file is registered with hipfile, hipfile creates duplicates of the file descriptor. One buffered, one unbuffered (O_DIRECT). hipFile uses these internal file descriptors when performing IO.

### DIFF
--- a/hipfile/src/amd_detail/file.cpp
+++ b/hipfile/src/amd_detail/file.cpp
@@ -56,6 +56,18 @@ UnregisteredFile::getFd() const noexcept
     return m_fd;
 }
 
+FileDescriptor &&
+UnregisteredFile::takeFdBuffered(const PassKey<File> &) noexcept
+{
+    return std::move(m_fd_buffered);
+}
+
+FileDescriptor &&
+UnregisteredFile::takeFdUnbuffered(const PassKey<File> &) noexcept
+{
+    return std::move(m_fd_unbuffered);
+}
+
 struct statx
 UnregisteredFile::getStatx() const noexcept
 {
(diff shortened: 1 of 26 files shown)


hipfile/src/amd_detail/file.cpp
```diff
@@ -35,8 +36,18 @@ UnregisteredFile::UnregisteredFile(int fd)
...
+      m_fd_unbuffered(FileDescriptor(Context<Sys>::get()->fcntl(fd, F_DUPFD_CLOEXEC, 0)))
 {
+    if (m_flags & O_DIRECT) {
+        Context<Sys>::get()->fcntl(m_fd_buffered.get(), F_SETFL,
+                                   static_cast<unsigned long>(m_flags & ~O_DIRECT));
+    }
+    else {
+        Context<Sys>::get()->fcntl(m_fd_unbuffered.get(), F_SETFL,
+                                   static_cast<unsigned long>(m_flags | O_DIRECT));
+    }
```
jordan-turbofish, line 50: I think you need to reopen the file to get the behavior you want here.  duping gives you another descriptor that ends up pointing at the same `struct file`, so the flags are shared.

kurtmcmillan: You are correct. I'll fix that.

kurtmcmillan: I've addressed this in #134

kurtmcmillan:
Closing in favor of #134

PR #134, opened 2025-12-15
hipfile: Ensure hipfile has unbuffered and buffered file descriptors

kurtmcmillan:
## Motivation

No longer require files to have the O_DIRECT flag set when they are registered with hipFile. This change will enable hipFile to eventually handle unaligned IO.

## Technical Details

When a file is registered with hipfile, duplicate the file descriptor and set/unset the O_DIRECT flag on the duplicate.

### DIFF
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -86,7 +86,8 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
         try {
             switch (io_type) {
                 case IoType::Read:
-                    io_bytes = Context<Sys>::get()->pread(file->getFd(), bounce_buffer.get(), count, offset);
+                    io_bytes =
+                        Context<Sys>::get()->pread(file->getClientFd(), bounce_buffer.get(), count, offset);
                     if (io_bytes > 0) {
                         Context<Hip>::get()->hipMemcpy(device_buffer_position, bounce_buffer.get(),
                                                        static_cast<size_t>(io_bytes), hipMemcpyHostToDevice);
@@ -96,7 +97,8 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
                     Context<Hip>::get()->hipMemcpy(bounce_buffer.get(), device_buffer_position, count,
                                                    hipMemcpyDeviceToHost);
                     Context<Hip>::get()->hipStreamSynchronize(nullptr);
-                    io_bytes = Context<Sys>::get()->pwrite(file->getFd(), bounce_buffer.get(), count, offset);
+                    io_bytes =
+                        Context<Sys>::get()->pwrite(file->getClientFd(), bounce_buffer.get(), count, offset);
                     break;
                 default:
                     throw std::runtime_error("Invalid IO type");
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -86,8 +86,8 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
         try {
             switch (io_type) {
                 case IoType::Read:
-                    io_bytes = Context<Sys>::get()->pread(file->getUnbufferedFd(), bounce_buffer.get(), count,
-                                                          offset);
+                    io_bytes =
+                        Context<Sys>::get()->pread(file->getBufferedFd(), bounce_buffer.get(), count, offset);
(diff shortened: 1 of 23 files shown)


src/amd_detail/backend/fallback.cpp
```diff
@@ -86,8 +86,8 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
         try {
             switch (io_type) {
                 case IoType::Read:
-                    io_bytes =
-                        Context<Sys>::get()->pread(file->getClientFd(), bounce_buffer.get(), count, offset);
+                    io_bytes = Context<Sys>::get()->pread(file->getUnbufferedFd(), bounce_buffer.get(), count,
```
jordan-turbofish: I think you want the buffered file descriptor for fallback, right?

src/amd_detail/file.cpp
```diff
@@ -38,16 +39,20 @@ UnregisteredFile::UnregisteredFile(int fd)
...
+        buffered_fd   = FileDescriptor::make_managed(
+            Context<Sys>::get()->open(path.c_str(), (flags | O_CLOEXEC) & ~O_DIRECT));
     }
     else {
         buffered_fd   = FileDescriptor::make_unmanaged(fd);
-        unbuffered_fd = FileDescriptor::make_managed(Context<Sys>::get()->fcntl(fd, F_DUPFD_CLOEXEC, 0));
-        Context<Sys>::get()->fcntl(unbuffered_fd.get(), F_SETFL,
-                                   static_cast<unsigned long>(flags | O_DIRECT));
+        unbuffered_fd = FileDescriptor::make_managed(
+            Context<Sys>::get()->open(path.c_str(), (flags | O_CLOEXEC) | O_DIRECT));
```
jordan-turbofish: Do we need to read the symlink first?  I believe the open will follow the symlink and I think the kernel has special handling around /proc/self/fd so you can still open it that way after it has been deleted.

src/amd_detail/file.cpp
```diff
@@ -37,6 +38,17 @@ UnregisteredFile::UnregisteredFile(int fd)
...
+    if (flags & O_DIRECT) {
+        unbuffered_fd = FileDescriptor::make_unmanaged(fd);
+        buffered_fd   = FileDescriptor::make_managed(Context<Sys>::get()->fcntl(fd, F_DUPFD_CLOEXEC, 0));
+        Context<Sys>::get()->fcntl(buffered_fd.get(), F_SETFL, static_cast<unsigned long>(flags & ~O_DIRECT));
+    }
+    else {
+        buffered_fd   = FileDescriptor::make_unmanaged(fd);
+        unbuffered_fd = FileDescriptor::make_managed(Context<Sys>::get()->fcntl(fd, F_DUPFD_CLOEXEC, 0));
+        Context<Sys>::get()->fcntl(unbuffered_fd.get(), F_SETFL,
+                                   static_cast<unsigned long>(flags | O_DIRECT));
```
jordan-turbofish: You could address this in a future PR, but I think unbuffered_fd would have to be optional to support filesystems without O_DIRECT support.

src/amd_detail/backend/fallback.cpp
kurtmcmillan: After our discussion yesterday I have updated the fallback path to use the buffered fd. I also added a call to fdatasync after the write finishes to ensure that a following O_DIRECT read will see the write.

src/amd_detail/file.cpp
kurtmcmillan: Updated to open /proc/self/fd/## directly.

src/amd_detail/file.cpp
kurtmcmillan: I'll do this in my next PR.

src/amd_detail/file.cpp
```diff
@@ -25,52 +25,27 @@ using std::vector;
...
 hipFileHandle_t
 IFile::getHandle() const
 {
     return reinterpret_cast<hipFileHandle_t>(const_cast<IFile *>(this));
 }
 
-File::File(const UnregisteredFile &uf, const PassKey<FileMap> &)
-    : fd{uf.getFd()}, stx{uf.getStatx()}, status_flags{uf.getFlags()}, mountinfo{uf.getMountInfo()}
+File::File(UnregisteredFile &&uf, const PassKey<FileMap> &)
+    : fd{uf.fd}, stx{uf.stx}, status_flags{uf.flags}, mountinfo{uf.mountinfo}
```
jordan-turbofish: Should this invalidate the UnregisteredFile?

jordan-turbofish (approved):

PR #143, opened 2025-12-17
hipfile: Handle filesystems without O_DIRECT support

kurtmcmillan:
Allow files to be registered even if opening the file with O_DIRECT fails with EINVAL. When hipfile cannot create a unbuffered file descriptor IO to the file will use the fallback/POSIX path.

### DIFF
--- a/src/amd_detail/file.cpp
+++ b/src/amd_detail/file.cpp
@@ -47,9 +47,10 @@ UnregisteredFile::UnregisteredFile(int fd)
                 Context<Sys>::get()->open(path.c_str(), (flags | O_CLOEXEC) | O_DIRECT));
         }
         catch (const std::system_error &e) {
-            if (e.code().value() == EINVAL) {
-                unbuffered_fd = nullopt;
+            if (e.code().value() != EINVAL) {
+                throw;
             }
+            unbuffered_fd = nullopt;
         }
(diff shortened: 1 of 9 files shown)


src/amd_detail/file.cpp
```diff
@@ -40,9 +41,16 @@ UnregisteredFile::UnregisteredFile(int fd)
...
-            Context<Sys>::get()->open(path.c_str(), (flags | O_CLOEXEC) | O_DIRECT));
+        buffered_fd = FileDescriptor::make_unmanaged(fd);
+        try {
+            unbuffered_fd = FileDescriptor::make_managed(
+                Context<Sys>::get()->open(path.c_str(), (flags | O_CLOEXEC) | O_DIRECT));
+        }
+        catch (const std::system_error &e) {
+            if (e.code().value() == EINVAL) {
+                unbuffered_fd = nullopt;
+            }
```
jordan-turbofish, line 52: Should this rethrow if it wasn't an EINVAL error?

jordan-turbofish (approved):

src/amd_detail/file.cpp
kurtmcmillan: Good catch. Fixed.

PR #144, opened 2025-12-19
CI: Fix CMake config to produce a "runtime" package.

riley-dixon:
`rocm_create_package` will not produce a runtime package unless the `BUILD_SHARED_LIBS` option is enabled.

This PR fixes that, and then uploads the runtime package as a build artifact as well.

AIROCFILE-97

### DIFF
--- a/cmake/AISInstall.cmake
+++ b/cmake/AISInstall.cmake
@@ -68,6 +68,10 @@ set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_SOURCE_DIR}/LICENSE.md")
 set(CPACK_RPM_PACKAGE_LICENSE "MIT")
 
 # Create the package
+# rocm_create_package makes decisions on what type of package to produce
+# based on BUILD_SHARED_LIBS. If we don't set this here we will only get
+# a dev package.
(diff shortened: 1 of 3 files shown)


cmake/AISInstall.cmake
```diff
@@ -68,6 +68,10 @@ set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_SOURCE_DIR}/LICENSE.md")
 set(CPACK_RPM_PACKAGE_LICENSE "MIT")
 
 # Create the package
+# rocm_create_package makes decisions on what type of package to produce
+# based on BUILD_SHARED_LIBS. If we don't set this here we will only get
+# a dev package.
+set(BUILD_SHARED_LIBS ON)
```
derobins: I feel like this should be moved to the root CMakeLists.txt file. It's more of a broad library construction thing than just a packaging thing.

riley-dixon: Was thinking the same thing but couldn't find a suitable location for it as we just build everything right now. I've moved it under the "Options" header, but feel free to make further changes if you would like.

derobins (approved):

PR #148, opened 2026-01-02
CI: Avoid recompiling hipFile in later stages of CI.

riley-dixon:
Replaces #141.

This PR removes redundantly compiling hipFile during later stages in the CI workflow, namely for system tests.

Where possible, the installed hipFile library from the DEB/RPM package should be used. This is already the case for FIO. However, for system tests executed via CTest we currently need to reuse the entire build directory. Removing this build-time dependency would be ideal, but is out of scope for this PR.

This PR also sets the hipFile package install path to the non-versioned ROCm directory (`/opt/rocm`).

AIROCFILE-97

### DIFF
--- a/.github/workflows/build-ais.yml
+++ b/.github/workflows/build-ais.yml
@@ -427,29 +427,6 @@ jobs:
           docker cp \
             ${GITHUB_WORKSPACE}/hipfile-build-dir/. \
             "${AIS_CONTAINER_NAME}:/ais/hipFile/build"
-      - name: Generate build files for hipFile targeting the AMD platform (amdclang++)
-        run: |
-          docker exec \
-            -t \
-            -w /ais/hipFile/build \
-            ${AIS_CONTAINER_NAME} \
-            /bin/bash -c '
-              cmake \
-                -DCMAKE_CXX_COMPILER=amdclang++ \
-                -DCMAKE_CXX_FLAGS="-Werror" \
-                -DCMAKE_HIP_PLATFORM=amd \
-                -DAIS_BUILD_DOCS=ON \
-                ..
-            '
-      - name: Build hipFile for the AMD platform (amdclang++)
-        run: |
-          docker exec \
-            -t \
-            -w /ais/hipFile/build \
-            ${AIS_CONTAINER_NAME} \
-            /bin/bash -c '
-              cmake --build . --parallel
-            '
       - name: Run hipFile system tests for the AMD platform (amdclang++)
         run: |
           docker exec \
--- a/.github/workflows/build-ais.yml
+++ b/.github/workflows/build-ais.yml
@@ -220,6 +220,22 @@ jobs:
           if-no-files-found: error
           retention-days: 1
           compression-level: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ais.yml
```diff
@@ -414,6 +414,19 @@ jobs:
...
               }}
             '
+      # GitHub Build Artifacts do not preserve file permissions, namely the executable bit.
+      - name: Fix file permissions on GTest executables
+        run: |
+          chmod a+x \
+            ${GITHUB_WORKSPACE}/hipfile-build-dir/test/hipfile_system_tests \
+            ${GITHUB_WORKSPACE}/hipfile-build-dir/test/amd_detail/batch_mt \
+            ${GITHUB_WORKSPACE}/hipfile-build-dir/test/amd_detail/state_mt \
+            ${GITHUB_WORKSPACE}/hipfile-build-dir/examples/aiscp/aiscp
```
jordan-turbofish, line 424: I know that it would be an archive inside of an archive, but would it be worth creating a tar archive for the build directory, so you don't have this list to maintain?

derobins (approved):

CMakeLists.txt
```diff
@@ -117,6 +117,17 @@ list(APPEND CMAKE_PREFIX_PATH
...
+# Note: CMAKE_INSTALL_PREFIX is set to a default by project().
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(CMAKE_INSTALL_PREFIX "${ROCM_PATH}" CACHE PATH "The path where hipFile should be installed" FORCE)
+endif()
+
+# Fix library install directory to "lib" to be inline with the rest of ROCm
+# Note: If testing is enabled, installing GTest calls GNUInstallDirs which will
+#       set this before ROCMInstallTargets can set it.
+set(CMAKE_INSTALL_LIBDIR "lib" CACHE STRING "Directory name for installed ROCm libraries")
+
```
jordan-turbofish, line 130: Would it be worth setting CMAKE_INSTALL_PREFIX on the configuration instead of changing the default?  Also, when I test this, the `ROCM_PATH` is set to /opt/rocm, which is a symlink to the path of the versioned rocm directory.  I think the package should be installing to the real path.

jordan-turbofish (commented):
A couple comments.

.github/workflows/build-ais.yml
riley-dixon: If this list changes frequently then yes I think that would be worthwhile until we can figure out how to produce a "test" package compatible with CTest. Not too worried about it at this time though.

CMakeLists.txt
riley-dixon: In TheRock, `ROCM_PATH` (or more typically `ROCM_DIR`) is set by a driver build script that gets passed into CMake - which eventually hits a block like above. So, I think the thing that needs to change instead is how ROCM_PATH gets set. Right now, our CI isn't setup to be configurable on a variable ROCm version.

>  I think the package should be installing to the real path.

I agree but point back to the limitation that currently exists in our CI.

PR #151, opened 2026-01-08
Rename hipFileOpStatusError()

derobins:
To hipFileGetOpErrorString() to better align with HIP, which uses hipGetErrorString() with hipError_t values.


kurtmcmillan:
I thought we were supposed to have the mapping cuFileXYZ -> hipFileXYZ. Should we define both `hipFileGetOpErrorString()` and  `hipFileOpStatusError()` to ensure we preserve the mapping? Should `hipFileGetOpErrorString()` be `hipFileExtGetOpErrorString()`

derobins:
> I thought we were supposed to have the mapping cuFileXYZ -> hipFileXYZ. Should we define both `hipFileGetOpErrorString()` and `hipFileOpStatusError()` to ensure we preserve the mapping? Should `hipFileGetOpErrorString()` be `hipFileExtGetOpErrorString()`

The cuFile API call is an oddball: `cufileop_status_error()`, so it doesn't fit the normal cuFile pattern.

derobins:
The failures are due to fio, which I can fix immediately after this goes in.

riley-dixon (approved):
CI should be fixed soon as a change in a different repo is needed in conjunction with this change.

PR #156, opened 2026-01-15
CI: Create a nightly release workflow

riley-dixon:
## Motivation

We would like to produce a nightly release of hipFile.

## Technical Details

Uses GH CLI for creating the release and uploading the packages to GitHub. There used to be a 1st party action to do this but has since been deprecated a few years ago.

Uses a glob pattern to fetch and reference the built hipFile packages. Best effort is used to only match runtime and development libraries, but we will need to verify this as we produce additional packages. This glob pattern will need to be updated/tested if a new platform is added to the support matrix or if the package filename changes.

## Test Plan

Does this pass the existing build workflow and are packages uploaded to the nightly release?

## Test Result

Ran the CI on my personal fork while the new nightly workflow had an added trigger of running on a `pull_request`.
https://github.com/riley-dixon/hipFile-test/actions/runs/21042360478?pr=14

AIHIPFILE-97

### DIFF
--- a/.github/workflows/build-ais.yml
+++ b/.github/workflows/build-ais.yml
@@ -10,6 +10,7 @@ env:
       inputs.platform == 'ubuntu' && 'apt'
     }}
   AIS_PKG_TYPE: ${{ inputs.platform == 'ubuntu' && 'DEB' || 'RPM' }}
+  AIS_ROCM_VERSION: 6.4.2
   # Code coverage report only vetted to work for amdclang++ on Ubuntu
   AIS_USE_CODE_COVERAGE: ${{ inputs.cxx_compiler == 'amdclang++' && inputs.platform == 'ubuntu' }}
 on:
@@ -100,6 +101,7 @@ jobs:
       - name: Generate build files for hipFile targeting the AMD platform (${{ inputs.cxx_compiler }})
         run: |
           docker exec \
+            -e "_AIS_ROCM_VERSION=${AIS_ROCM_VERSION}" \
             -t \
             -w /ais/hipFile/build \
             ${AIS_CONTAINER_NAME} \
@@ -110,6 +112,7 @@ jobs:
                 -DCMAKE_HIP_PLATFORM=amd \
                 -DAIS_BUILD_DOCS=ON \
                 -DAIS_USE_CODE_COVERAGE=${{ env.AIS_USE_CODE_COVERAGE == 'true' && 'ON' || 'OFF' }} \
+                -DROCM_PATH="/opt/rocm-${_AIS_ROCM_VERSION}" \
                 ..
             '
       - name: Build hipFile for the AMD platform (${{ inputs.cxx_compiler }})
--- a/.github/workflows/nightly-build-ais.yml
+++ b/.github/workflows/nightly-build-ais.yml
@@ -1,0 +1,34 @@
+name: AIS_nightly_build
+run-name: Create a nightly build of hipFile
+on:
+  workflow_call:
+    inputs:
+      platform:
+        required: true
+        type: string
+    outputs:
+      ais_hipfile_pkg_dev_filename:
+        description: "Filename for the hipFile development DEB/RPM package."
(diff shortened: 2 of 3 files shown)


jordan-turbofish (commented):
Do we want to change the installation path for the packages before releasing them nightly?  Right now, files from the packages will be orphaned if the ROCm version is upgraded.

I think that we should update the ROCm versions in the containers so we're building the package against the latest release.

riley-dixon:
> Do we want to change the installation path for the packages before releasing them nightly? Right now, files from the packages will be orphaned if the ROCm version is upgraded.

@jordan-turbofish I did make a note regarding this concern that will be posted with the release. Regarding the fix, this should be a separate effort outside of this PR. Our CI does not currently have the ability to be variable on ROCm releases. I do plan to work on that next.

I think the question then becomes, do we want this PR to wait for that to be completed first, or have the next PR enabling CI to switch between ROCm versions update the nightly release at the same time.

> I think that we should update the ROCm versions in the containers so we're building the package against the latest release.

Sure, we can do that separately.

jordan-turbofish:
> @jordan-turbofish I did make a note regarding this concern that will be posted with the release. Regarding the fix, this should be a separate effort outside of this PR. Our CI does not currently have the ability to be variable on ROCm releases. I do plan to work on that next.

Do we want to spend time on building packages for multiple ROCm versions, or is releasing a package for the latest version good enough? If we updated ROCm in the Docker images, this small change would have our package install to the versioned path:

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8be557e..0552b59 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,7 +120,8 @@ list(APPEND CMAKE_PREFIX_PATH
 # Set hipFile Install Path to the ROCm directory
 # Note: CMAKE_INSTALL_PREFIX is set to a default by project().
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-    set(CMAKE_INSTALL_PREFIX "${ROCM_PATH}" CACHE PATH "The path where hipFile should be installed" FORCE)
+    file(REAL_PATH "${ROCM_PATH}" ROCM_ABSOLUTE_PATH)
+    set(CMAKE_INSTALL_PREFIX "${ROCM_ABSOLUTE_PATH}" CACHE PATH "The path where hipFile should be installed" FORCE)
 endif()

 # Fix library install directory to "lib" to be inline with the rest of ROCm
```

jordan-turbofish (approved):

derobins:
> > @jordan-turbofish I did make a note regarding this concern that will be posted with the release. Regarding the fix, this should be a separate effort outside of this PR. Our CI does not currently have the ability to be variable on ROCm releases. I do plan to work on that next.
> 
> Do we want to spend time on building packages for multiple ROCm versions, or is releasing a package for the latest version good enough? If we updated ROCm in the Docker images, this small change would have our package install to the versioned path:
> 
> ```diff
> diff --git a/CMakeLists.txt b/CMakeLists.txt
> index 8be557e..0552b59 100644
> --- a/CMakeLists.txt
> +++ b/CMakeLists.txt
> @@ -120,7 +120,8 @@ list(APPEND CMAKE_PREFIX_PATH
>  # Set hipFile Install Path to the ROCm directory
>  # Note: CMAKE_INSTALL_PREFIX is set to a default by project().
>  if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
> -    set(CMAKE_INSTALL_PREFIX "${ROCM_PATH}" CACHE PATH "The path where hipFile should be installed" FORCE)
> +    file(REAL_PATH "${ROCM_PATH}" ROCM_ABSOLUTE_PATH)
> +    set(CMAKE_INSTALL_PREFIX "${ROCM_ABSOLUTE_PATH}" CACHE PATH "The path where hipFile should be installed" FORCE)
>  endif()
> 
>  # Fix library install directory to "lib" to be inline with the rest of ROCm
> ```

I think right now the latest version is fine.

PR #157, opened 2026-01-15
aiscp: Handle files > 2,147,479,552 bytes

kurtmcmillan:
This change allows aiscp to copy files that are larger than the maximum allowed IO size on linux.


kurtmcmillan:
Going to re-work this implementation and then put it back up for review.

PR #160, opened 2026-01-19
CI: Resolve CodeQL concerns

riley-dixon:
This resolves some CodeQL signals that could be considered false positives since our CI variables are not provided by external workflows. There concerns in particular are regarding potential shell injection should our callable workflows ever take in input from an external user. It recommends storing inputs in the environment rather than referencing them directly within a Bash step. We also encapsulate these instances within quotes to further try to prevent injections via substitutions.

AIHIPFILE-97

### DIFF
--- a/.github/workflows/build-ais-ci-image.yml
+++ b/.github/workflows/build-ais-ci-image.yml
@@ -1,6 +1,7 @@
 name: new_CI_image
 run-name: Build the Docker image for AIS CI.
 env:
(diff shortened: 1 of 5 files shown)


riley-dixon:
PR#162 was used to test that there was no adverse change made to the build_ais_ci_image workflow.

Workflow run: https://github.com/ROCm/hipFile/actions/runs/21154308647?pr=160

derobins (approved):

PR #162, opened 2026-01-19
Rildixon/test pr 160

riley-dixon:
Test PR #160 - We want to test that the changes made here will not break the workflow to update the CI image. However, that PR comes from a fork and cannot push images into GHCR. This PR will not replace 160.


riley-dixon:
The new CI image was built successfully. Closing this PR.

PR #163, opened 2026-01-20
Add fallback async IO support.

jordan-turbofish:
This PR adds fallback async IO support to hipFile.  IO going through this path uses a pinned CPU bounce buffer to copy data between the disk and GPU memory.  Multiple operations will be queued on the stream in order to carry out the IO.

Read:
1. async_io_bind_params (if needed) will bind the current value of user-provided unfixed parameters.
2. async_io_cpu_copy will copy the data from the disk to the CPU bounce buffer.
3. hipFileMemcpyKernel will copy the data from the CPU bounce buffer to GPU memory.
4. async_io_cleanup will clean up memory used by the IO operation an set the bytes_transferred value.

Write:
1. async_io_bind_params (if needed) will bind the current value of user-provided unfixed parameters.
2. hipFileMemcpyKernel will copy the data from the GPU memory to the CPU bounce buffer.
3. async_io_cpu_copy will copy the data from the CPU bounce buffer to the disk.
4. async_io_cleanup will clean up memory used by the IO operation an set the bytes_transferred value.

AIHIPFILE-67

### DIFF
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -122,7 +122,9 @@ Fallback::async_io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBu
                    hoff_t *file_offset_p, hoff_t *buffer_offset_p, ssize_t *bytes_transferred_p,
                    std::shared_ptr<IStream> stream)
 {
-    if (!paramsValid(buffer, *size_p, *file_offset_p, *buffer_offset_p)) {
+    size_t limited_size = min(*size_p, hipFile::MAX_RW_COUNT);
+
+    if (!paramsValid(buffer, limited_size, *file_offset_p, *buffer_offset_p)) {
         throw std::invalid_argument("The selected file or buffer region is invalid");
     }
 
@@ -199,7 +201,7 @@ async_io_bind_params(void *userargs)
     const hoff_t *file_offset = get_variant_ptr(op->file_offset);
     op->file_offset.emplace<const hoff_t>(*file_offset);
     const size_t *size = get_variant_ptr(op->size);
-    op->size           = *size;
+    op->size           = std::min(*size, hipFile::MAX_RW_COUNT);
 
     if (std::get<size_t>(op->size) > op->submitted_size) {
         op->bytes_transferred_internal = -hipFileInvalidValue;
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -221,7 +221,7 @@ async_io_cleanup(void *userargs)
     try {
         Context<AsyncMonitor>::get()->completeOp(op);
     }
-    catch (std::invalid_argument &e) {
+    catch (const std::invalid_argument &) {
         *bytes_transferred = -hipFileInternalError;
         return;
     }
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -239,12 +239,12 @@ async_io_cpu_copy(void *userargs)
         return;
     }
 
-    try {
-        while (bytes_transferred < size) {
-            void *cur_buf_position = reinterpret_cast<void *>(
-                reinterpret_cast<uintptr_t>(op->bounceBufferHostPtr()) + bytes_transferred);
-            hoff_t cur_file_offset = file_offset + static_cast<hoff_t>(bytes_transferred);
-            size_t remaining_bytes = size - bytes_transferred;
+    while (bytes_transferred < size) {
+        void *cur_buf_position = reinterpret_cast<void *>(
+            reinterpret_cast<uintptr_t>(op->bounceBufferHostPtr()) + bytes_transferred);
+        hoff_t cur_file_offset = file_offset + static_cast<hoff_t>(bytes_transferred);
+        size_t remaining_bytes = size - bytes_transferred;
+        try {
             switch (op->io_type) {
                 case IoType::Read:
                     ret = Context<Sys>::get()->pread(op->file->getBufferedFd(), cur_buf_position,
@@ -257,18 +257,19 @@ async_io_cpu_copy(void *userargs)
                 default:
                     throw std::runtime_error("Invalid IO type");
             }
-            if (ret == 0) {
-                break;
-            }
-            bytes_transferred += static_cast<size_t>(ret);
         }
-        op->bytes_transferred_internal = static_cast<ssize_t>(bytes_transferred);
+        catch (const std::system_error &e) {
+            if (e.code().value() == EINTR) {
+                continue;
+            }
+            op->bytes_transferred_internal = -1;
+            return;
+        }
+        if (ret == 0) {
+            break;
+        }
+        bytes_transferred += static_cast<size_t>(ret);
     }
-    catch (std::system_error &e) {
-        op->bytes_transferred_internal = -1;
-    }
-    catch (Hip::RuntimeError &e) {
-        op->bytes_transferred_internal = -hipFileDriverError;
-    }
+    op->bytes_transferred_internal = static_cast<ssize_t>(bytes_transferred);
 }
 }
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -60,24 +60,10 @@ ssize_t
 Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer, size_t size,
              hoff_t file_offset, hoff_t buffer_offset, size_t chunk_size)
 {
-    size_t buflen{buffer->getLength()};
-
     size = min(size, hipFile::MAX_RW_COUNT);
 
-    if (file_offset < 0) {
-        throw std::invalid_argument("Negative file offset");
-    }
-
-    if (buffer_offset < 0) {
-        throw std::invalid_argument("Negative buffer offset");
-    }
-
-    if (buflen <= static_cast<size_t>(buffer_offset)) {
-        throw std::invalid_argument("Buffer offset larger than buffer length");
-    }
-
-    if (buflen - static_cast<size_t>(buffer_offset) < size) {
-        throw std::invalid_argument("IO could overflow buffer");
+    if (!paramsValid(buffer, size, file_offset, buffer_offset)) {
+        throw std::invalid_argument("The selected file or buffer region is invalid");
     }
 
     auto ptr     = Context<Sys>::get()->mmap(nullptr, chunk_size, PROT_READ | PROT_WRITE,
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -3,23 +3,32 @@
  * SPDX-License-Identifier: MIT
  */
 
+#include "async.h"
 #include "backend.h"
 #include "buffer.h"
+#include "backend/asyncop-fallback.h"
+#include "backend/memcpy-kernel.h"
 #include "context.h"
 #include "fallback.h"
 #include "file.h"
 #include "hip.h"
+#include "hipfile.h"
 #include "io.h"
 #include "sys.h"
+#include "stream.h"
+#include "util.h"
 
 #include <algorithm>
 #include <cerrno>
 #include <cstdint>
 #include <hip/hip_runtime_api.h>
 #include <hip/driver_types.h>
+#include <memory>
 #include <stdexcept>
 #include <sys/mman.h>
+#include <syslog.h>
 #include <system_error>
+#include <variant>
 
 using namespace hipFile;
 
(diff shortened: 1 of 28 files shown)


src/amd_detail/stream.cpp
```diff
@@ -67,7 +75,9 @@ StreamMap::registerStream(hipStream_t hip_stream, uint32_t flags)
         throw std::invalid_argument("Stream is already registered");
     }
 
-    auto stream = std::shared_ptr<Stream>(new Stream(hip_stream, flags, PassKey<StreamMap>{}));
+    hipDevice_t device_id = Context<Hip>::get()->hipStreamGetDevice(hip_stream);
```
kurtmcmillan: Minor: Would it be better to have the constructor of Stream to initialize `device_id` using `Context<Hip>::get()->hipStreamGetDevice(hip_stream)` instead of passing it in?

src/amd_detail/backend/memcpy-kernel.hip
```diff
@@ -0,0 +1,103 @@
...
+    const size_t  num_chunks        = size / sizeof(size_t);
+
+    switch (io_type) {
+        case IoType::Read:
+            for (size_t idx = start_idx; idx < num_chunks; idx += blockDim.x) {
+                gpu_buffer_region[idx] = cpu_buffer[idx];
+            }
+            break;
+        case IoType::Write:
+            for (size_t idx = start_idx; idx < num_chunks; idx += blockDim.x) {
```
riley-dixon, line 71: Do we need to consider `blockIdx.x` (or probably more correctly `gridDim.x`?) when incrementing? Or are we expecting to only run one block at a time?

src/amd_detail/io.cpp
```diff
@@ -0,0 +1,30 @@
...
+paramsValid(const std::shared_ptr<IBuffer> &buffer, size_t size, hoff_t file_offset, hoff_t buffer_offset)
+{
+    size_t buffer_length = buffer->getLength();
+    if (file_offset < 0) {
+        return false;
+    }
+    if (buffer_offset < 0) {
+        return false;
+    }
+    if (buffer_length < static_cast<size_t>(buffer_offset)) {
```
kurtmcmillan: Isn't buffer_length == buffer_offset also bad?

src/amd_detail/io.cpp
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+#include "buffer.h"
+#include "io.h"
+
+namespace hipFile {
+
+bool
+paramsValid(const std::shared_ptr<IBuffer> &buffer, size_t size, hoff_t file_offset, hoff_t buffer_offset)
```
kurtmcmillan, line 12: This matches the checks done in fastpath::io and fallback::io. Could you update those methods to call paramsValid

src/amd_detail/backend/memcpy-kernel.hip
riley-dixon: Okay I see on the next commit that we hardcode `gridDim` to 1. Maybe worthwhile then removing `blockIdx.x` from `start_idx` and adding a comment that we only support running on a single block for now?

src/amd_detail/backend/fallback.cpp
```diff
@@ -121,3 +130,159 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
...
+            Context<Sys>::get()->syslog(LOG_CRIT,
+                                        "Unable to enqueue async cleanup function. This will leak memory.");
+        }
+        throw;
+    }
+}
+
+extern "C" {
+void
+async_io_bind_params(void *userargs)
```
riley-dixon, line 195: Let's say at least one of `op->stream->fixed*` is true, and at least one is false so that this function gets called. How do the fixed parameters not get overwritten? Or is this not something that we enforce?

jordan-turbofish: Each of the parameters is stored in the variant of form <T, T*>.  If T is stored in the variant, get_variant_ptr returns a pointer to that value.  So, it will overwrite itself with its own value.  It T* is stored in the variant, get_variant_ptr returns the stored pointer, and will use the value it points to to update the variant.

src/amd_detail/backend/fallback.cpp
```diff
@@ -121,3 +130,159 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
...
+                    throw std::runtime_error("Invalid IO type");
+            }
+            if (ret == 0) {
+                break;
+            }
+            bytes_transferred += static_cast<size_t>(ret);
+        }
+        op->bytes_transferred_internal = static_cast<ssize_t>(bytes_transferred);
+    }
+    catch (std::system_error &e) {
```
kurtmcmillan: Should we retry pread/pwrite on EINTR?

src/amd_detail/backend/fallback.cpp
```diff
@@ -121,3 +130,159 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
...
+    auto op = static_cast<AsyncOpFallback *>(userargs);
+    // Bind params. Will maintain same value if already bound.
+    const hoff_t *buffer_offset = get_variant_ptr(op->buffer_offset);
+    op->buffer_offset.emplace<const hoff_t>(*buffer_offset);
+    const hoff_t *file_offset = get_variant_ptr(op->file_offset);
+    op->file_offset.emplace<const hoff_t>(*file_offset);
+    const size_t *size = get_variant_ptr(op->size);
+    op->size           = *size;
+
+    if (std::get<size_t>(op->size) > op->submitted_size) {
```
kurtmcmillan, line 206: Should we limit the IO size to `hipFile::MAX_RW_COUNT` here? Or is that done elsewhere?

src/amd_detail/backend/fallback.cpp
```diff
@@ -121,3 +130,159 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
...
+    }
+
+    *bytes_transferred_p = 0;
+
+    if (*size_p == 0) {
+        return;
+    }
+
+    auto op = std::shared_ptr<AsyncOpFallback>(new AsyncOpFallback(
+        type, file, buffer, stream, size_p, file_offset_p, buffer_offset_p, bytes_transferred_p));
```
kurtmcmillan, line 142: in the cufile docs they say that if the IO size is not known at submission that size should be set to the maximum allowed size (I don't know what that is). When AsyncOpFallback is initialized it allocates the bounce buffer. If the maximum allowed IO size is large, wouldn't that be problematic.

test/amd_detail/async.cpp
```diff
@@ -281,4 +289,381 @@ TEST_F(HipFileAsyncMonitor, addOp_without_completeOp_prints_error_on_AsyncMonito
...
+    EXPECT_CALL(*mbuffer, getLength).WillOnce(Return(1024 * 1024));
+    EXPECT_THROW(Fallback().async_io(io_type, mfile, mbuffer, &size, &file_offset, &buffer_offset,
+                                     &bytes_written, mstream),
+                 std::invalid_argument);
+}
+
+TEST_P(FallbackAsyncIO, bufferOffsetTooLargeReturnsError)
+{
+    buffer_offset = 1024 * 1024 + 1;
+    EXPECT_CALL(*mbuffer, getLength).WillOnce(Return(1024 * 1024));
```
kurtmcmillan: could do:

```
buffer_offset = 1024 * 1024;
EXPECT_CALL(*mbuffer, getLength).WillOnce(Return(buffer_offset));
```

test/system/async.cpp
```diff
@@ -0,0 +1,872 @@
...
+              HipFileOpError(hipFileInvalidValue));
+}
+
+TEST_P(HipAsyncReadWriteStreamFixedWithParams, nullBytesReadReturnsError)
+{
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, nullptr, stream),
+              HipFileOpError(hipFileInvalidValue));
+}
+
+TEST_P(HipAsyncReadWriteStreamFixedWithParams, ioSizeTooLargeForBuffer)
```
kurtmcmillan, line 502: What does nvidia do here? Do they truncate the IO size to the size of the buffer?

test/system/async.cpp
```diff
@@ -0,0 +1,872 @@
...
+#ifdef __HIP_PLATFORM_AMD__
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, &bytes_transferred, stream),
+              HipFileOpError(hipFileInvalidValue));
+#else
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, &bytes_transferred, stream),
+              HIPFILE_SUCCESS);
+#endif
+}
+
+TEST_P(HipAsyncReadWriteStreamFixedWithParams, fileOffsetNegative)
```
kurtmcmillan, line 517: With the negative file offset or negative buffer offset what is NVIDIA doing? Ignoring the value?

test/system/async.cpp
```diff
@@ -0,0 +1,872 @@
...
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, &bytes_transferred, stream),
+              HipFileOpError(hipFileInvalidValue));
+#else
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, &bytes_transferred, stream),
+              HIPFILE_SUCCESS);
+#endif
+}
+
+TEST_P(HipAsyncReadWriteStreamFixedWithParams, bufferOffsetTooLarge)
+{
```
kurtmcmillan, line 548: ??? Maybe we should document the behavior differences when it comes to validation of parameters in async io. I wonder these same differences exist in the sync apis

test/system/async.cpp
```diff
@@ -0,0 +1,872 @@
...
+
+TEST_P(HipAsyncReadWriteParamsUnchanged, ioSizeUnaligned)
+{
+    io_size -= 1;
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, &bytes_transferred, stream),
+              HIPFILE_SUCCESS);
+    ASSERT_EQ(hipStreamSynchronize(stream), hipSuccess);
+    ASSERT_EQ(bytes_transferred, io_size);
+    assertFileAndMemoryRegionsMatch(dev_ptr, buffer_offset, tf.fd, file_offset, io_size);
+    if (name == "hipFileReadAsync") {
```
kurtmcmillan, line 620: Wouldn't the following zeroed region checks fail if if `buffer_offset` or `file_offset` were non-zero.

kurtmcmillan: or is that what `ReadWriteParamsUnchanged` implies?

test/system/async.cpp
```diff
@@ -0,0 +1,872 @@
...
+    bytes_transferred = 1;
+    ASSERT_EQ(io_op(fh, dev_ptr, &io_size, &file_offset, &buffer_offset, &bytes_transferred, stream),
+              HIPFILE_SUCCESS);
+    ASSERT_EQ(bytes_transferred, 0);
+}
+
+TEST_P(HipAsyncReadWriteParamsUnchanged, multipleOpsOnSameStreamAreSequential)
+{
+
+    size_t num_ios = (io_size / 8_KiB) - 1;
```
kurtmcmillan, line 674: Why are you overlapping IO here?

kurtmcmillan (approved):

src/amd_detail/backend/fallback.cpp
riley-dixon: Ah okay, thanks for the explanation. I went back to the AsyncOp ctor and that helped solidify how which type gets originally assigned to determine which is "fixed" (and as you mention overwritten by itself).

riley-dixon (approved):

src/amd_detail/backend/fallback.cpp
jordan-turbofish: Maybe I'm interpreting it wrong, but I read that as the user specifies the maximum that the IO could be based on how their program works, rather than a max constant.

test/system/async.cpp
jordan-turbofish: Yeah, the parameters for these tests do not change after submission.  I'm only varying the stream flags.

test/system/async.cpp
jordan-turbofish: I am trying to test that the ios are sequential within a stream.  So the IOs look like this (A & B representing 8KiB sections of the source region).

```
┌───┬───┐
│ A │ B │
└───┴───┘
    ┌───┬───┐
    │ A │ B │
    └───┴───┘
        ┌───┬───┐
        │ A │ B │
        └───┴───┘

```

If the operations are sequential, the result should be:

```
┌───┬───┬───┬───┐
│ A │ A │ A │ B │
└───┴───┴───┴───┘
```

If any earlier IO happened after a later IO, you'd get something like this:

```
┌───┬───┬───┬───┐
│ A │ B │ A │ B │
└───┴───┴───┴───┘
```

test/system/async.cpp
jordan-turbofish: On NVIDIA, the call itself is successful, then they return the error on the bytes_transferred variable.  I've added comments and updated the NVIDIA tests to show this.  At the moment, our errors don't all match (although I don't think we should match the hipFileInternalError cuFile returns form the one test).  I'm currently returning hipFileInvalidValue for parameter issues.

riley-dixon (approved):
Reiterating approval (in case you were wanting a re-review from myself).

PR #164, opened 2026-01-20
CI: Update CI images to use ROCm 7.1.1

riley-dixon:
Updates ROCm version used in the CI container to 7.1.1

AMDGPU repo removed as we don't need any packages from that repository (mostly kernel packages).


riley-dixon:
Hold off on merging this until #156 gets checked in.

jordan-turbofish (approved):

PR #165, opened 2026-01-20
hipFile: Limit IO size to MAX_RW_COUNT on fastpath

kurtmcmillan:
## Motivation

When testing aiscp with files larger than MAX_RW_COUNT amdgpu/kfd would return an ILLEGAL SEEK error. To address this, the AIS path (fastpath) will the limit IO size to MAX_RW_COUNT. This matches the behavior of POSIX read/write (see man pages).

aiscp is also updated to handle files larger than MAX_RW_COUNT.

Finally, since I was in the area, aiscp now handles files with zero size.

AIHIPFILE-111

### DIFF
--- a/examples/aiscp/aiscp.cpp
+++ b/examples/aiscp/aiscp.cpp
@@ -106,6 +106,7 @@ main(int argc, char *argv[])
     }
 
     if (0 == file_size) {
+        exit_status = EXIT_SUCCESS;
         goto close_dst;
     }
 
--- a/examples/aiscp/aiscp.cpp
+++ b/examples/aiscp/aiscp.cpp
@@ -24,7 +24,7 @@
 #include <unistd.h>
 
 /// @brief AISCP_CHUNK_SIZE defines the maximum size of each aiscp IO operation.
-/// Currently AISCP_CHUNKS_SIZE is defined to match the maximum IO size on Linux
+/// Currently AISCP_CHUNK_SIZE is defined to match the maximum IO size on Linux
 /// (MAX_RW_COUNT).
 #ifndef AISCP_CHUNK_SIZE
 #define AISCP_CHUNK_SIZE 0x7ffff000LU
--- a/examples/aiscp/aiscp.cpp
+++ b/examples/aiscp/aiscp.cpp
@@ -88,6 +88,15 @@ alignUp(size_t value, size_t align)
     return (value + align - 1) & ~(align - 1);
 }
 
+/// @brief Determines if value is a power of two
+/// @param value The value to inspect
+/// @return True if value is a power of two, false otherwise
+static inline bool
+is_power_of_two(size_t value)
+{
+    return (value > 0) && ((value & (value - 1)) == 0);
+}
+
 int
 main(int argc, char *argv[])
 {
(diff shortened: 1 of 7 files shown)


examples/aiscp/aiscp.cpp
```diff
@@ -106,42 +124,47 @@ main(int argc, char *argv[])
...
+            goto free_devbuf;
+        }
 
-    if (file_size < buffer_size) {
-        if (-1 == ftruncate(dst_fd, static_cast<off_t>(file_size))) {
-            fprintf(stderr, "Could not truncate %s (%zu) (%s)\n", dst_path, file_size, strerror(errno));
+        nwrite = 0;
+        while (nwrite < nread) {
+            nbytes =
+                hipFileWrite(dst_handle, devbuf, alignUp(static_cast<size_t>(nread - nwrite), block_size),
```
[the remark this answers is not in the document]
kurtmcmillan: The point here is to write block size chunks and then truncate the file to match the original file size. If we don't write block size chunks then hipfile cannot use the AIS path.

jordan-turbofish (approved):

src/amd_detail/backend/fastpath.cpp
```diff
@@ -177,6 +177,12 @@ Fastpath::io(IoType type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer, si
         throw std::invalid_argument("IO could overflow buffer");
     }
 
+    // Currently, when IO sizes > MAX_RW_COUNT are submitted to amdgpu/kfd an
+    // Illegal Seek error is returned. To avoid this, hipFile limits IO size to
+    // MAX_RW_COUNT. When amdgpu/kfd properly handles IO sizes > MAX_RW_COUNT
+    // this can be removed.
+    size = std::min(size, MAX_RW_COUNT);
```
derobins, line 184: Should update the Doxygen in hipfile.h to indicate that this limit exists and the number of IO bytes not matching the requested number of bytes is an expected condition for large writes.

derobins: Also need to update the CHANGELOG.md file

examples/aiscp/aiscp.cpp
derobins, line 128: I would probably either check that `st_blksize` is a power of 2 or at least comment that it must be a power of 2 for aiscp to function with IO sizes that are not a power of 2 given that it will be used as the `align` param for `alignUp()`. I don't think we'll encounter that in practice and this is just a sample program, but it's good to be thorough.

examples/aiscp/aiscp.cpp
```diff
@@ -106,42 +124,47 @@ main(int argc, char *argv[])
...
-
-    nbytes = hipFileWrite(dst_handle, devbuf, buffer_size, 0, 0);
-    if (nbytes < 0 || buffer_size != static_cast<size_t>(nbytes)) {
-        fprintf(stderr, "Could not write to %s (%zd) (%s)\n", src_path, nbytes,
-                IS_HIPFILE_ERR(nbytes) ? HIPFILE_ERRSTR(nbytes) : strerror(errno));
-        goto free_devbuf;
-    }
+    // Copy the file chunk-by-chunk until we hit EOF
+    do {
+        nread = hipFileRead(src_handle, devbuf, buffer_size, ncopy, 0);
```
derobins: `ncopy` is of type `ssize_t` and the parameter is an offset of type `hoff_t`, which is confusing. I would probably rename `ncopy` and use the correct type.

kurtmcmillan: Fixed.

examples/aiscp/aiscp.cpp
kurtmcmillan: I don't think so either, but I've added a power of two check just in case.

src/amd_detail/backend/fastpath.cpp
kurtmcmillan: Documented in doxygen in hipfile.h and in CHANGELOG.md. Unsure if I did the changelog correctly.

derobins (approved):

PR #167, opened 2026-01-22
CI: Fix build_AIS workflow for non-PR triggers.

riley-dixon:
build_AIS was written with the assumption only PRs would trigger the workflow to be run. This is no longer the case with our nightly release workflow. As such, ${AIS_PR_NUMBER} resolves to an empty string which makes ${AIS_CONTAINER_NAME} invalid as it starts with a '_' character.

To fix this, we will now provide a fallback when ${AIS_PR_NUMBER} is empty. This is set the workflow Run ID which is unique to the triggering event. If the workflow is re-run, this ID is reused. To cover this, we add a concurrency tag to only allow one single instance of the nightly release workflow to run at a time.

I'll test this by trying to trigger a manual workflow.

### DIFF
--- a/.github/workflows/build-ais.yml
+++ b/.github/workflows/build-ais.yml
@@ -65,8 +65,9 @@ jobs:
           echo "AIS_SAFE_COMPILER_NAME=$(echo ${AIS_INPUT_CXX_COMPILER} | sed 's|[\+]|plus|g')" >> "${GITHUB_ENV}"
       - name: Set AIS CI container name
         # We should expect that there are multiple instances of this job with different cxx_compilers.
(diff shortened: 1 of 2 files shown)


riley-dixon:
Tested on personal fork using a workflow_dispatch trigger: https://github.com/riley-dixon/hipFile-test/actions/runs/21261421377/job/61189620640

${AIS_CONTAINER_NAME} is now prefixed with the run ID instead of the PR #. Pull_request jobs are unchanged.

jordan-turbofish (approved):

derobins (approved):

PR #181, opened 2026-02-05
Move util/aischeck to tools/ais-check/ais-check

derobins:


cmake/AISInstall.cmake
```diff
@@ -17,8 +17,8 @@ rocm_install(
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
 
-# Install utilities
-install(PROGRAMS util/aischeck DESTINATION bin)
+# Install AIS tools
+install(PROGRAMS tools/ais-check/ais-check DESTINATION bin)
```
gaoikawa, line 21: Do we want this double 'ais-check/ais-check' folder layout?  Or is ais-check the one file (and only ever one file?) does it need it's own directory?  Not too picky either way.  Thanks.

gaoikawa (approved):

cmake/AISInstall.cmake
derobins: I suspect there will probably be other tools and it'd be useful to keep them separated, as they may have other files, READMEs, etc. so dumping them all in one directory might make a mess.

gaoikawa: Works for me!  Thx.

PR #184, opened 2026-02-12
Use make_shared to create a shared pointer

jbelloncastro:
## Motivation

Small code simplification.

## Technical Details

Constructing a `shared_ptr` instance with `make_shared` is simpler and more efficient than using the raw pointer constructor.

Shared pointers need extra storage for the atomic reference count. `make_shared` has the opportunity to reserve the memory for both ref-count and the object itself in one operation, while the alternative requires two separate allocations, with the ref-count storage requiring an extra indirection to object itself.

## Test Plan

The change has no effect on functionality so all the tests should have the same outcome.

## Test Result

N/A

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -138,8 +138,8 @@ Fallback::async_io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBu
         return;
     }
 
-    auto op = std::shared_ptr<AsyncOpFallback>(new AsyncOpFallback(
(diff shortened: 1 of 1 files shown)


jordan-turbofish:
Hi @jbelloncastro.  Thanks for having a look at improving hipFile.  I agree that make_shared is better in general, but in this case we override new to allocate pinned host memory.  make_shared will call the default allocator, which will not give us the allocation type we want.

PR #189, opened 2026-02-18
Statistics reporting tool: ais-stats

zbyrne:
## Motivation

[AIHIPFILE-53] Infinity Storage as a storage solution must have sufficient performance counters & statistics for proper debugging, system component workflow understanding, and performance troubleshooting.

## Technical Details

Counters are stored in a shared-memory-backed mmapped buffer as an array of atomic_uint64_ts. The library uses unix sockets and control messages to communicate the fd for the buffer with the reporting app.

## Test Plan

Unit tests were added for some simple behaviors. Additionally, the example program aiscp was launched by ais-stats and IO stats were reported.

## Test Result

$ ctest . -R 'HipFileStats*'
Test project /home/AMD/zbyrne/git/hipFile/build
    Start 412: HipFileStats.statsAddFastPathRead
1/6 Test #412: HipFileStats.statsAddFastPathRead ........   Passed    0.02 sec
    Start 413: HipFileStats.statsAddFastPathWrite
2/6 Test #413: HipFileStats.statsAddFastPathWrite .......   Passed    0.02 sec
    Start 414: HipFileStats.statsAddFallbackPathRead
3/6 Test #414: HipFileStats.statsAddFallbackPathRead ....   Passed    0.02 sec
    Start 415: HipFileStats.statsAddFallbackPathWrite
4/6 Test #415: HipFileStats.statsAddFallbackPathWrite ...   Passed    0.02 sec
    Start 416: HipFileStats.StatsServerLifetime
5/6 Test #416: HipFileStats.StatsServerLifetime .........   Passed    0.02 sec
    Start 417: HipFileStats.GenerateReportV1
6/6 Test #417: HipFileStats.GenerateReportV1 ............   Passed    0.02 sec

100% tests passed, 0 tests failed out of 6

$ build/tools/ais-stats/ais-stats build/examples/aiscp/aiscp README.md /home/AMD/zbyrne/README.md
Total fast path reads (B): 0
Total fast path writes (B):0
Total fallback path reads (B): 1729
Total fallback path writes (B): 524288

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

[AIHIPFILE-53]: https://amd-hub.atlassian.net/browse/AIHIPFILE-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/amd_detail/backend/fallback.cpp
+++ b/src/amd_detail/backend/fallback.cpp
@@ -14,6 +14,7 @@
 #include "hip.h"
 #include "hipfile.h"
 #include "io.h"
+#include "stats.h"
 #include "sys.h"
 #include "stream.h"
 #include "util.h"
@@ -114,6 +115,16 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
         }
     } while (static_cast<size_t>(total_io_bytes) < size);
 
+    switch (io_type) {
+        case IoType::Read:
+            statsAddFallbackPathRead(static_cast<size_t>(total_io_bytes));
+            break;
+        case IoType::Write:
+            statsAddFallbackPathWrite(static_cast<size_t>(total_io_bytes));
+            break;
+        default:
+            break;
+    }
     return total_io_bytes;
 }
 
--- a/src/amd_detail/backend/fastpath.cpp
+++ b/src/amd_detail/backend/fastpath.cpp
@@ -10,6 +10,7 @@
 #include "hip.h"
 #include "hipfile.h"
 #include "io.h"
+#include "stats.h"
 
 #include <cstdint>
 #include <fcntl.h>
@@ -183,6 +184,15 @@ Fastpath::io(IoType type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer, si
         default:
             throw std::runtime_error("Invalid IoType");
     }
-
+    switch (type) {
+        case IoType::Read:
+            statsAddFastPathRead(nbytes);
+            break;
+        case IoType::Write:
+            statsAddFastPathWrite(nbytes);
+            break;
+        default:
+            break;
+    }
     return static_cast<ssize_t>(nbytes);
 }
--- a/src/amd_detail/stats.cpp
+++ b/src/amd_detail/stats.cpp
@@ -47,6 +47,39 @@ sendFd(int sock, int fd) noexcept
     return 0;
 }
 
+static int
+recvFd(int sockfd) noexcept
+{
+    int      data, fd;
+    iovec    iov{&data, sizeof(data)};
+    msghdr   msgh;
+    cmsghdr *cmsgp;
+
+    union {
+        char    buff[CMSG_SPACE(sizeof(int))];
+        cmsghdr align;
+    } controlMsg;
+
+    msgh.msg_name       = nullptr;
+    msgh.msg_namelen    = 0;
+    msgh.msg_iov        = &iov;
+    msgh.msg_iovlen     = 1;
+    msgh.msg_control    = controlMsg.buff;
+    msgh.msg_controllen = sizeof(controlMsg.buff);
+
+    if (recvmsg(sockfd, &msgh, 0) == -1)
+        return -1;
+
+    cmsgp = CMSG_FIRSTHDR(&msgh);
+    if (cmsgp == NULL || cmsgp->cmsg_len != CMSG_LEN(sizeof(int)) || cmsgp->cmsg_level != SOL_SOCKET ||
+        cmsgp->cmsg_type != SCM_RIGHTS) {
+        errno = EINVAL;
+        return -1;
+    }
+    memcpy(&fd, CMSG_DATA(cmsgp), sizeof(int));
+    return fd;
+}
+
 static void
 populateSocketAddr(sockaddr_un &addr, pid_t pid) noexcept
 {
@@ -121,6 +154,85 @@ StatsServer::threadFn()
         }
     }
     close(sock);
+}
+
+StatsClient::StatsClient(pid_t p)
+    : m_pfd{FileDescriptor::make_managed(Context<Sys>::get()->pidfd_open(p, 0))}, m_pid{p}
+{
+}
+
+bool
+StatsClient::pollProcess(int timeout)
+{
+    if (m_pfd.get() == -1) {
+        return true;
+    }
+    pollfd pfd{m_pfd.get(), POLLIN, 0};
+    return poll(&pfd, 1, timeout) > 0;
+}
+
+bool
+StatsClient::connectServer()
+{
+    FileDescriptor sock{FileDescriptor::make_managed(socket(AF_UNIX, SOCK_STREAM, 0))};
+    if (sock.get() == -1) {
+        return false;
+    }
+    int success{-1};
+
+    sockaddr_un addr;
+    populateSocketAddr(addr, m_pid);
+    for (int timeout{0}; !pollProcess(timeout); ++timeout) { // backoff on connect attempts
+        success = connect(sock.get(), reinterpret_cast<sockaddr *>(&addr), sizeof(struct sockaddr_un));
+        if (success == 0) {
+            break;
+        }
+    }
+    if (success == -1) {
+        return false;
+    }
+    m_sfd = FileDescriptor::make_managed(recvFd(sock.get()));
+    return true;
+}
+
+bool
+StatsClient::generateReport(std::ostream &stream)
+{
+    if (m_sfd.get() == -1) {
+        return false;
+    }
+    void *shm = mmap(nullptr, sizeof(Stats), PROT_READ, MAP_SHARED, m_sfd.get(), 0);
+    if (shm == reinterpret_cast<void *>(-1)) {
+        return false;
+    }
+    Stats *stats = reinterpret_cast<Stats *>(shm);
+    if (stats == nullptr) {
+        return false;
+    }
+    switch (stats->version) {
+        case 1:
+            generateReportV1(stream, stats);
+            break;
+        default:
+            break;
+    }
+    munmap(stats, sizeof(Stats));
+    return true;
+}
+
+void
+StatsClient::generateReportV1(std::ostream &stream, const Stats *stats)
+{
+    if (stats == nullptr) {
+        return;
+    }
+    stream << "Total fast path reads (B): " << stats->getCounter(StatsCounters::TotalFastPathReadBytes).load()
+           << "\nTotal fast path writes (B): "
+           << stats->getCounter(StatsCounters::TotalFastPathWriteBytes).load()
+           << "\nTotal fallback path reads (B): "
+           << stats->getCounter(StatsCounters::TotalFallbackPathReadBytes).load()
+           << "\nTotal fallback path writes (B): "
+           << stats->getCounter(StatsCounters::TotalFallbackPathWriteBytes).load() << '\n';
 }
 
 void
--- a/src/amd_detail/stats.cpp
+++ b/src/amd_detail/stats.cpp
@@ -1,0 +1,53 @@
+/* Copyright (c) Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include "context.h"
+#include "stats.h"
+
+namespace hipFile {
+StatsServer::StatsServer()
+{
+}
+
+StatsServer::~StatsServer()
+{
+}
+
+void
+statsAddFastPathRead(uint64_t bytes)
+{
+    Stats *stats{Context<StatsServer>::get()->getStats()};
+    if (stats) {
+        stats->getCounter(StatsCounters::TotalFastPathReadBytes) += bytes;
+    }
+}
+
+void
+statsAddFastPathWrite(uint64_t bytes)
+{
+    Stats *stats{Context<StatsServer>::get()->getStats()};
+    if (stats) {
+        stats->getCounter(StatsCounters::TotalFastPathWriteBytes) += bytes;
+    }
+}
+
+void
+statsAddFallbackPathRead(uint64_t bytes)
+{
+    Stats *stats{Context<StatsServer>::get()->getStats()};
+    if (stats) {
+        stats->getCounter(StatsCounters::TotalFallbackPathReadBytes) += bytes;
+    }
+}
+
+void
+statsAddFallbackPathWrite(uint64_t bytes)
+{
+    Stats *stats{Context<StatsServer>::get()->getStats()};
+    if (stats) {
+        stats->getCounter(StatsCounters::TotalFallbackPathWriteBytes) += bytes;
+    }
+}
+}
--- a/src/amd_detail/stats.cpp
+++ b/src/amd_detail/stats.cpp
@@ -5,14 +5,36 @@
 
 #include "context.h"
 #include "stats.h"
+#include "sys.h"
+
+#include <fcntl.h>
+#include <unistd.h>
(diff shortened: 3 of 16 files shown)


src/amd_detail/stats.cpp
```diff
@@ -0,0 +1,269 @@
...
+    m_thread  = std::thread(&StatsServer::threadFn, this);
+}
+
+StatsServer::~StatsServer()
+{
+    if (m_thread.joinable()) {
+        m_run = false;
+        m_thread.join();
+    }
+    if (m_stats) {
```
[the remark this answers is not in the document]
zbyrne: the idea is the contents of the Stats should persist since ais-stats has to read it after the library closes, we don't want to call the dtor.

src/amd_detail/stats.h
```diff
@@ -0,0 +1,45 @@
...
+    TotalFallbackPathWriteBytes,
+
+    Capacity = 64,
+};
+
+struct Stats {
+    using Array = std::array<std::atomic_uint64_t, static_cast<std::size_t>(StatsCounters::Capacity)>;
+    const uint64_t version{1};
+    Array          counters;
+};
```
kurtmcmillan, line 43: Would it be better to have the counters be fields of the stats struct rather than being elements in an array?

eg.

Version 1:

```
struct stats {
  const uint64_t version{1};
  // Version 1 Fields
  uint64_t fastpath_read_bytes;
  uint64_t fastpath_write_bytes;
}

```
Version 2:

```
struct stats {
  const uint64_t version{2};
  // Version 1 Fields
  uint64_t fastpath_read_bytes;
  uint64_t fastpath_write_bytes;
  // Version 2 Fields
  uint64_t fastpath_read_err_bytes;
  uint64_t fastpath_read_count;
  uint64_t fastpath_read_err_count;
  ...
}
```

src/amd_detail/stats.cpp
```diff
@@ -5,14 +5,29 @@
...
+#include <sys/mman.h>
 
 namespace hipFile {
-StatsServer::StatsServer()
+StatsServer::StatsServer() : m_fd{}
 {
+    m_fd = FileDescriptor::make_managed(Context<Sys>::get()->memfd_create("AISSTATS", MFD_ALLOW_SEALING));
+    int fd{m_fd.get()};
+    Context<Sys>::get()->ftruncate(fd, sizeof(Stats));
+    Context<Sys>::get()->fcntl(fd, F_ADD_SEALS, F_SEAL_SHRINK);
```
kurtmcmillan: Can you also apply the write seal to ensure that future mappings are not allowed to write to the mapping?

src/amd_detail/stats.h
zbyrne: I had them like that originally, I changed it to this, thinking it would make more sense to version the index enum

src/amd_detail/stats.cpp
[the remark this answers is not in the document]
kurtmcmillan: The mapping lives as long as one process has a handle to the mapping. So it is safe to destroy the mapping. It should also be safe to call the destructor unless the destructor is zeroing the data.

test/amd_detail/stats.cpp
```diff
@@ -49,4 +50,14 @@ STAT_TEST(FastPathWrite)
 STAT_TEST(FallbackPathRead)
 STAT_TEST(FallbackPathWrite)
 
+TEST_F(HipFileStats, StatsServerLifetime)
+{
+    MSys msys{};
```
kurtmcmillan: We typically use strict mocks to make the test fail if we miss mocking methods.

`StrictMock<MSys> msys{};`

src/amd_detail/stats.cpp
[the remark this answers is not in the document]
zbyrne: yeah, i'm going to move the whole thing into a unique_ptr with a custom deleter to call munmap, i'll add the dtor there

src/amd_detail/stats.cpp
```diff
@@ -20,16 +69,43 @@ StatsServer::StatsServer() : m_fd{}
...
+StatsServer::threadFn()
+{
+    int   sock = socket(AF_UNIX, SOCK_STREAM, 0);
+    pid_t pid  = getpid();
+
+    sockaddr_un addr;
+    populateSocketAddr(addr, pid);
+    bind(sock, reinterpret_cast<const sockaddr *>(&addr), sizeof(addr));
+    listen(sock, 64);
+    while (m_run) {
```
kurtmcmillan: Could the server thread could get stuck in poll and not ever see m_run become false? I've seen others use an event fd  for this.

Thread A polls on both the socket and the event fd.
Thread B writes false to m_run and then writes to the event fd
Thread A wakes up out of poll, sees that event fd is signaled or m_run is false and then exits.

src/amd_detail/stats.h
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return m_stats;
+    }
+
+private:
+    void           threadFn();
+    Stats         *m_stats{nullptr};
+    FileDescriptor m_fd;
+    std::thread    m_thread;
+    bool           m_run{true};
```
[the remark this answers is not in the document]
kurtmcmillan: If you use the eventfd you might not even need m_run. Having data on the eventfd would be the signal to the thread that its time to exit.

src/amd_detail/stats.cpp
zbyrne: Yeah, I can use an event instead, I think it's safe the way it is, the timeout sent to poll is 1ms, so it shouldn't get stuck.

src/amd_detail/backend/fastpath.cpp
```diff
@@ -183,6 +184,15 @@ Fastpath::io(IoType type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer, si
         default:
             throw std::runtime_error("Invalid IoType");
     }
-
+    switch (type) {
```
sbates130272, line 187: @zbyrne do we understand any performance implications of adding this call on the fastpath? Are there any changes for performance regression?

sbates130272 (approved):
@zbyrne some minor comments. Looks good to me!

src/amd_detail/backend/fastpath.cpp
zbyrne: @sbates130272 I haven't looked too deeply into the perf cost. Once the pr is closer to done I'll do an A/B with fio. If the call overhead is noticeable I can inline and streamline a bit, if I start seeing contention on the atomic adds with lots of jobs... I dunno. I'm thinking we should add a config value to enable or disable collection anyway, IIRC gds-stats has 3 config levels with increasing levels of detail collected.

src/amd_detail/stats.cpp
zbyrne: I don't think I can add write seal to this, the man page says "trying  to  create  new  shared,
              writable memory-mappings via mmap(2) will also fail with EPERM" and "Using  the  F_ADD_SEALS  operation  to  set the F_SEAL_WRITE seal fails with EBUSY if any writable, shared mapping exists"

src/amd_detail/stats.cpp
```diff
@@ -0,0 +1,269 @@
...
+    memset(&addr, 0, sizeof(addr));
+    addr.sun_family  = AF_UNIX;
+    addr.sun_path[0] = '\0'; // abstract namespace
+    snprintf(&(addr.sun_path[1]), sizeof(addr.sun_path) - 1, "AISSTATS%08x", static_cast<uint32_t>(pid));
+}
+
+namespace hipFile {
+StatsServer::StatsServer() : m_fd{}
+{
+    m_fd = FileDescriptor::make_managed(Context<Sys>::get()->memfd_create("AISSTATS", MFD_ALLOW_SEALING));
```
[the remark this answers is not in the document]
zbyrne: CI passed, mman.h seems to work on all supported platforms.

test/amd_detail/stats.cpp
riley-dixon, line 1: Are we able to move MStateServer into a header file so that other units can mock out the stats functions in unit tests?

src/amd_detail/stats.h
```diff
@@ -0,0 +1,83 @@
...
+    const std::atomic_uint64_t &getCounter(StatsCounters index) const
+    {
+        return counters[static_cast<std::size_t>(index)];
+    }
+};
+
+void statsAddFastPathRead(uint64_t bytes);
+void statsAddFastPathWrite(uint64_t bytes);
+void statsAddFallbackPathRead(uint64_t bytes);
+void statsAddFallbackPathWrite(uint64_t bytes);
```
riley-dixon, line 48: GTest doesn't work on free functions for mocking/patching. Are we able to move them into the Stats struct?

src/amd_detail/stats.h
```diff
@@ -0,0 +1,83 @@
...
+
+class StatsServer {
```
riley-dixon, line 50: Can you please create an abstract interface (e.g. IStatsServer) for GTest compatibility.

src/amd_detail/stats.h
```diff
@@ -0,0 +1,83 @@
...
+
+private:
+    void           threadFn();
+    FileDescriptor m_fd;
+    FileDescriptor m_efd;
+    UniqueStats    m_stats;
+    std::thread    m_thread;
+};
+
+class StatsClient {
```
riley-dixon, line 70: StatsClient seems to only be used in ais-stats so I'm not as worried about setting this class up for GTest mocking, but something to consider if in the future we see the hipFile library using this class.

src/amd_detail/backend/fallback.cpp
```diff
@@ -114,6 +115,16 @@ Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer,
         }
     } while (static_cast<size_t>(total_io_bytes) < size);
 
+    switch (io_type) {
```
riley-dixon, line 118: If we have an error thrown in the middle of a large IO, do we want to track the IO stats that have been processed up to this point?

src/amd_detail/stats.h
riley-dixon: On further consideration, this might require a bit of a redesign. Fallback & Fastpath would both need to have a StatsServer instance assuming these functions get placed into it.

Having getStats() patched to return null might be good enough to at least limit how much stats code gets executed in unit tests (at least for how the functions are written today).

src/amd_detail/backend/fallback.cpp
zbyrne: probably, the current counters are really just simple proof-of-concept to make sure all the plumbing is working. There's a doc in confluence for planning more detailed data gathering (still in draft)

src/amd_detail/stats.h
zbyrne: pretty much everything in StatsClient started out as part of ais-stats, I moved it because I wanted to keep sendFd() and recvFd() together.

src/amd_detail/stats.h
zbyrne: Just so MStatsServer isn't extending StatsServer?

src/amd_detail/stats.h
zbyrne: I'm currently using our Context<> pattern as a lazy singleton. A single StatsServer could probably be part of the DriverState

src/amd_detail/stats.h
riley-dixon: Exactly that :)

src/amd_detail/backend/fallback.cpp
riley-dixon: Sounds good then. If this is then out-of-scope and will be done later we can resolve this comment.

src/amd_detail/stats.h
zbyrne: This seems to anger the ContextOverride
`/home/AMD/zbyrne/git/hipFile/src/amd_detail/context.h: In instantiation of ‘hipFile::ContextOverride<T>::ContextOverride(U*) [with U = hipFile::MStatsServer; T = hipFile::StatsServer]’:
/home/AMD/zbyrne/git/hipFile/test/amd_detail/stats.cpp:20:22:   required from here
/home/AMD/zbyrne/git/hipFile/src/amd_detail/context.h:84:30: error: cannot convert ‘hipFile::MStatsServer*’ to ‘hipFile::StatsServer*’`

src/amd_detail/stats.h
riley-dixon: Sorry I might have had a wrong understanding with how the IO Backends were needing to access the StatsServer.

I think that might work, something like `Context<DriverState>::get()->statsAddFastPath(IOType)`? That would allow the function to keep the check to see if stat collection was enabled without needing to do that check in `io()`. Based on the current stats implementation, we wouldn't need to obtain the DriverState lock?

src/amd_detail/stats.h
riley-dixon: Ah sorry, my mistake. Please ignore, this isn't applicable since StatsServer access is managed via Context<>.

src/amd_detail/backend/fastpath.cpp
zbyrne: image

src/amd_detail/stats.h
zbyrne: @riley-dixon I think we decided this was ok on Friday, eh?

riley-dixon: It's not ideal but as long as the Context StatsServer is properly mocked and we aren't calling into any stats related stuff for IO unit tests it's fine.

src/amd_detail/stats.cpp
jordan-turbofish: Does setting F_SEAL_FUTURE_WRITE after the mmap work here?  I tried adding it and ais-stats tool still looks like its working.  From the man page:

```
Otherwise, behavior similar to  F_SEAL_WRITE  can  be  achieved  by  using
F_SEAL_FUTURE_WRITE,  which  will  prevent  future  writes via mmap(2) and
write(2) from succeeding while keeping existing shared writable mappings).
```

zbyrne: Yes, that looks like exactly what we want.

src/amd_detail/stats.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (stats) {
+        stats->getCounter(StatsCounters::TotalFallbackPathWriteBytes) += bytes;
+    }
+}
+
+HIPFILE_WARN_NO_GLOBAL_CTOR_OFF
+HIPFILE_WARN_NO_EXIT_DTOR_OFF
+static StatsServer *s_stats = Context<StatsServer>::get();
+HIPFILE_WARN_NO_EXIT_DTOR_ON
+HIPFILE_WARN_NO_GLOBAL_CTOR_ON
```
jordan-turbofish: Maybe I'm missing something, but I'm not sure what this is for.

Possibly related, if you want to control the static initialization order, you can add a `Context<StatsServer>::get();` to `src/amd_detail/context.cpp`.  I think you want it in between Sys and DriverState.

zbyrne: I just wanted it static initialized. I can move it into HipFileInit if you prefer.

jordan-turbofish: Okay.  Context<StatsServer>::get() has a function local static, so StatsServer is statically initialized the first time it is called.  It doesn't need to be assigned.  Adding the get call to HipFileInit will let you control the order of static initialization (which affects the destruction order).

jordan-turbofish (approved):

PR #198, opened 2026-02-27
Turn on the pip cache for Python GH actions

derobins:
The pylint and cmakelint actions use Python. This PR turns
on the pip cache.

### DIFF
--- a/.github/workflows/cmakelint.yml
+++ b/.github/workflows/cmakelint.yml
@@ -17,7 +17,8 @@ jobs:
 
       - name: Set up Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 #v6.2.0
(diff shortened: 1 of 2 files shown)


riley-dixon:
Are there other steps that need to be completed for caching to be correctly setup?
https://github.com/ROCm/hipFile/actions/runs/22500135457/job/65184788966#step:9:2

derobins:
> Are there other steps that need to be completed for caching to be correctly setup? https://github.com/ROCm/hipFile/actions/runs/22500135457/job/65184788966#step:9:2



> Are there other steps that need to be completed for caching to be correctly setup? https://github.com/ROCm/hipFile/actions/runs/22500135457/job/65184788966#step:9:2

I looked and I think it only caches if you use a requirements.txt file.

Closing

PR #199, opened 2026-02-27
Update INSTALL dependencies

derobins:


INSTALL.md
```diff
@@ -94,8 +94,12 @@ sudo systemctl reboot
 
 ### Prerequisites
 
-> [!NOTE]
-> hipFile relies on the unreleased [ROCm 7.2](https://github.com/ROCm/TheRock) libraries and associated [amdgpu](https://github.com/ROCm/amdgpu) drivers. We will update the install instructions when these are released.
+* CMake >= 3.21
+* C++ >= 17 (tested w/ clang++ & g++, we don't use GNU extensions)
+* ROCm >= 7.2
+* amdgpu >= 30.10.1
```
riley-dixon: 30.20.1 should be the minimum version. (Also agree with copilot about specifying dkms).

derobins: Will fix

INSTALL.md
```diff
@@ -80,7 +80,7 @@ If CMake can't find hipcc/nvcc, you can set `-DCMAKE_HIP_COMPILER=<path>`
 
 Supported platforms: Linux (Windows may be supported in a future release)
```
gaoikawa, line 82: Could you add something to the effect that in the linux kernel we require 'CONFIG_PCI_P2PDMA' to be enabled?  On by default in Ubuntu 24.04 (is this true?), but not in 22.x?  Can be done in Ubuntu22 via the hardware enablement kernel (hwe)?  (Need to confirm.)  Would be good if we could give some guidance on what has it enabled by default and what doesn't, and either a link to existing instructions for a few distros.

cc @jordan-turbofish

zbyrne (approved):

INSTALL.md
```diff
@@ -81,7 +81,7 @@ If CMake can't find hipcc/nvcc, you can set `-DCMAKE_HIP_COMPILER=<path>`
...
 
-Supported filesystems: Only ext4 is supported at this time
+Supported filesystems: Only ext4 and xfs are supported at this time
```
kurtmcmillan: We've only tested ext4 with the `data=ordered` mount option. `data` specifies `ext4`'s journaling mode. There are two other options. `ordered` is the default.

We should mention that here.

INSTALL.md
```diff
@@ -81,10 +81,12 @@ If CMake can't find hipcc/nvcc, you can set `-DCMAKE_HIP_COMPILER=<path>`
 
 Supported platforms: Linux (Windows may be supported in a future release)
 
-Supported filesystems: Only ext4 is supported at this time
+Supported filesystems: Only ext4 and xfs are supported at this time. ext4 only supports `data=ordered` mode.
```
kurtmcmillan: > Only ext4 and xfs are supported at this time. ext4 only supports `data=ordered` mode

It would probably read better if this was

> Only ext4* and xfs are supported at this time. ext4 must be mounted with the `data=ordered` mount option.

kurtmcmillan (approved):

PR #207, opened 2026-03-03
Add Stats Level config value for use with stats counter collection.

zbyrne:
## Motivation

AIHIPFILE-53 It's possible as we add more counters to ais-stats that we may start to see a performance overhead to the data collection. This setting will be used to let us control exactly how much data we collect from a given process in the event very detailed stats gathering has a large performance impact.

## Technical Details

Adds an environment variable(HIPFILE_STATS_LEVEL) and Configuration method(::statsLevel()) in the same vein as HIPFILE_FORCE_COMPAT_MODE.

## Test Plan

Added Environment and Configuration unit tests for the new functions.

## Test Result

All unit tests pass.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/src/amd_detail/configuration.cpp
+++ b/src/amd_detail/configuration.cpp
@@ -17,8 +17,6 @@ Configuration::Configuration() : m_fastpath(true), m_fallback(true), m_statsLeve
     if (maybe_env_force_compat && maybe_env_force_compat.value()) {
         m_fastpath = false;
     }
-    m_fastpath = !!getHipAmdFileReadPtr() && m_fastpath;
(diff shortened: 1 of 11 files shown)


gaoikawa:
It's early I realize and @derobins is re-working our docs, but we probably should add something to our public facing docs on stats collection 'soon'.  

hipFile -> docs -> (something related to stats, perhaps)  

Also, other libraries have a standalone environment variables explainer page, we should add something too.

Okay to capture that in another follow-on story/PR.

Thanks.

riley-dixon (approved):

PR #210, opened 2026-03-05
Update GTest CMake for use w/ sanitizers

derobins:
Ensure that GoogleTest is fetched and built from source with the same flags as the library when building with the sanitizers enabled. Without this, we don't see useful output when there are failures.

Part of AIHIPFILE-62

### DIFF
--- a/cmake/AISUseGTest.cmake
+++ b/cmake/AISUseGTest.cmake
@@ -7,24 +7,20 @@ include(FetchContent)
 # This line is used (or not) in FetchContent_Declare to determine
 # if we check for a system GoogleTest install first. When building
 # with the sanitizers, we HAVE to build GoogleTest from source.
-#
-# lint_cmake: -readability/wonkycase
 if(AIS_USE_SANITIZERS OR AIS_USE_INTEGER_SANITIZER OR AIS_USE_THREAD_SANITIZER)
-    FetchContent_Declare(
-      googletest
-      URL https://github.com/google/googletest/releases/download/v1.17.0/googletest-1.17.0.tar.gz
-      DOWNLOAD_EXTRACT_TIMESTAMP true
-      SYSTEM
-    )
+    set(AIS_LOCAL_GTEST_CHECK "")
 else()
-    FetchContent_Declare(
-      googletest
-      URL https://github.com/google/googletest/releases/download/v1.17.0/googletest-1.17.0.tar.gz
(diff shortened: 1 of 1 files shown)


cmake/AISUseGTest.cmake
```diff
@@ -4,25 +4,40 @@
...
+      URL https://github.com/google/googletest/releases/download/v1.17.0/googletest-1.17.0.tar.gz
+      DOWNLOAD_EXTRACT_TIMESTAMP true
+      SYSTEM
+    )
+else()
+    FetchContent_Declare(
+      googletest
+      URL https://github.com/google/googletest/releases/download/v1.17.0/googletest-1.17.0.tar.gz
+      DOWNLOAD_EXTRACT_TIMESTAMP true
+      FIND_PACKAGE_ARGS NAMES GTest         # <-- Different
```
jordan-turbofish: You could set a variable to these args or an empty string in the if statement and pass it into the FetchContent_Declare call to avoid duplication.

derobins:
@copilot please review

cmake/AISUseGTest.cmake
derobins: Yeah, I thought about that but copy-paste was easier than dealing with wonky string handling in CMake

jordan-turbofish: Lol.  Copilot made the same suggestion: https://github.com/ROCm/hipFile/pull/211/changes/67e7d9b9e3a34f50f3ebf333f3d928ae571c9507

Not sure why it opened a PR, though.

derobins: The problem is you get nonsense like this if you use an argument string:

CMake Error at /usr/share/cmake-3.28/Modules/FetchContent.cmake:1667 (message):
  CMake step for googletest failed: 1
Call Stack (most recent call first):
  /usr/share/cmake-3.28/Modules/FetchContent.cmake:1819:EVAL:2 (__FetchContent_directPopulate)
  /usr/share/cmake-3.28/Modules/FetchContent.cmake:1819 (cmake_language)
  /usr/share/cmake-3.28/Modules/FetchContent.cmake:2033 (FetchContent_Populate)
  cmake/AISUseGTest.cmake:30 (FetchContent_MakeAvailable)
  test/CMakeLists.txt:6 (include)

derobins: Fixed now

jordan-turbofish (approved):

PR #213, opened 2026-03-10
Rework sanitizers

derobins:
* Drop AIS_USE_INTEGER_SANITIZER and roll that into AIS_USE_SANITIZERS
* Add additional sanitizers:
    * float-divide-by-zero
    * vptr
    * local-bounds
    * nullability
* All sanitizers are clang-only (for now)
* Protect C++ compiler options with generator expressions
* Protect linker options with -Xarch_host
* Add the following linker options under sanitizers. These select the LLVM linker and runtime, but GNU unwind library.
    * -fuse-ld=lld
    * --rtlib=compiler-rt
    * --unwindlib=libgcc

Part of AIHIPFILE-62

### DIFF
--- a/cmake/AISCompilerOptions.cmake
+++ b/cmake/AISCompilerOptions.cmake
@@ -6,6 +6,7 @@
 
 include(AISClangCompilerOptions)
 include(AISGNUCompilerOptions)
+include(AISSanitizers)
 
 function(ais_set_compiler_flags target)
     get_target_property(sources ${target} SOURCES)
@@ -38,7 +39,9 @@ function(ais_set_compiler_flags target)
         endif()
     endforeach()
 
-    ais_add_sanitizers(${target})
+    if(AIS_USE_SANITIZERS OR AIS_USE_THREAD_SANITIZER)
+        ais_add_sanitizers(${target})
+    endif()
 
     if(NOT BUILD_TESTING)
         target_compile_options(${target} PRIVATE -fvisibility=hidden)
--- a/cmake/AISSanitizers.cmake
+++ b/cmake/AISSanitizers.cmake
@@ -2,53 +2,50 @@
 #
 # SPDX-License-Identifier: MIT
 
-option(AIS_USE_SANITIZERS "Build with -fsanitize=address, leak, and undefined" OFF)
-option(AIS_USE_THREAD_SANITIZER "Build with -fsanitize=thread (not compatible with AIS_USE_SANITIZERS)" OFF)
-option(AIS_USE_INTEGER_SANITIZER "Build with -fsanitize=integer (clang only)" OFF)
+option(AIS_USE_SANITIZERS "Build with sanitizers (clang only - see docs for a list)" OFF)
+option(AIS_USE_THREAD_SANITIZER "Build with -fsanitize=thread (clang only - not compatible with AIS_USE_SANITIZERS)" OFF)
 
 if(AIS_USE_THREAD_SANITIZER)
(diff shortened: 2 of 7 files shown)


cmake/AISSanitizers.cmake
```diff
@@ -2,53 +2,50 @@
...
     endif()
 
+    # Options for all sanitizers
+    target_compile_options(${target} PRIVATE
+        -fno-omit-frame-pointer
+    )
+    target_link_options(${target} PRIVATE
+        -fuse-ld=lld
+        --rtlib=compiler-rt
+        --unwindlib=libgcc
```
derobins, line 49: This uses llvm's ld and rtlib instead of the default GNU, but keeps GNU's unwinding, which is apparently considered more mature.

cmake/AISSanitizers.cmake
```diff
@@ -2,53 +2,50 @@
...
+option(AIS_USE_THREAD_SANITIZER "Build with -fsanitize=thread (clang only - not compatible with AIS_USE_SANITIZERS)" OFF)
 
 if(AIS_USE_THREAD_SANITIZER)
     message(WARNING "TSAN has known problems with higher levels of entropy, try using `sudo sysctl vm.mmap_rnd_bits=28` if you encounter errors concerning unexpected memory mappings.")
 endif()
 
 function(ais_add_sanitizers target)
+    if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+        message(FATAL_ERROR "The sanitizers are only available on clang")
+    endif()
```
derobins, line 15: Some of the options are clang-only, so I'm disabling GNU (for now). I'll rework this to support GNU in the future, but right now I want to get us clean on clang.

cmake/AISCompilerOptions.cmake
```diff
@@ -38,7 +39,9 @@ function(ais_set_compiler_flags target)
         endif()
     endforeach()
 
-    ais_add_sanitizers(${target})
+    if(AIS_USE_SANITIZERS OR AIS_USE_THREAD_SANITIZER)
+        ais_add_sanitizers(${target})
+    endif()
```
derobins, line 44: The check here avoids "are the sanitizers on?" checks in AISSanitizers

cmake/AISSanitizers.cmake
```diff
@@ -2,53 +2,50 @@
...
-        # TODO: Consider other sanitizer options
-        target_compile_options(${target} PRIVATE -fsanitize=address)
-        target_link_options(${target} PRIVATE -fsanitize=address)
-        target_compile_options(${target} PRIVATE -fsanitize=leak)
-        target_link_options(${target} PRIVATE -fsanitize=leak)
-        target_compile_options(${target} PRIVATE -fsanitize=undefined)
-        target_link_options(${target} PRIVATE -fsanitize=undefined)
-
-        target_compile_options(${target} PRIVATE -fno-omit-frame-pointer)
+        set(SANITIZER_LIST address,float-divide-by-zero,integer,leak,local-bounds,nullability,undefined,vptr)
```
derobins, line 22: All the sanitizers, because why not?

cmake/AISSanitizers.cmake
```diff
@@ -2,53 +2,50 @@
...
+        target_compile_options(${target} PRIVATE
+            $<$<COMPILE_LANGUAGE:CXX>:-fsanitize=${SANITIZER_LIST}>
+            $<$<COMPILE_LANGUAGE:CXX>:-fno-sanitize-recover=integer>
+        )
+        target_link_options(${target} PRIVATE
+            -Xarch_host -fsanitize=${SANITIZER_LIST}
+            -Xarch_host -fno-sanitize-recover=integer
```
derobins, line 29: Have to protect the options so they are C++/host only, otherwise we get warnings

zbyrne (approved):

PR #214, opened 2026-03-11
ais-check: Add a check for kernel P2PDMA support

derobins:
Looks for CONFIG_PCI_P2PDMA=y, m, or n in the config files

Also adds `uname -a`-like output

Part of AIHIPFILE-139

### DIFF
--- a/tools/ais-check/ais-check
+++ b/tools/ais-check/ais-check
@@ -38,7 +38,7 @@ def kernel_supports_p2pdma():
                         "CONFIG_PCI_P2PDMA=m"
                     ):
                         return True
-        except OSError as e:
+        except OSError:
             continue
 
     if not configs_found:
--- a/tools/ais-check/ais-check
+++ b/tools/ais-check/ais-check
@@ -22,20 +22,25 @@ def kernel_supports_p2pdma():
     Check for P2P DMA support in the kernel
     """
 
-    kr = os.uname().release
+    try:
+        kr = os.uname().release
 
-    for path, opener in [
-        (f"/boot/config-{kr}", open),
-        ("/proc/config.gz", gzip.open),
-    ]:
-        if os.path.exists(path):
-            with opener(path, "rt") as f:
-                for line in f:
-                    if line.startswith("CONFIG_PCI_P2PDMA=y") or line.startswith(
-                        "CONFIG_PCI_P2PDMA=m"
-                    ):
-                        return True
-    return False
+        for path, opener in [
+            (f"/boot/config-{kr}", open),
+            ("/proc/config.gz", gzip.open),
+        ]:
+            if os.path.exists(path):
+                with opener(path, "rt") as f:
+                    for line in f:
+                        if line.startswith("CONFIG_PCI_P2PDMA=y") or line.startswith(
+                            "CONFIG_PCI_P2PDMA=m"
+                        ):
+                            return True
+        return False
+
+    except (OSError, PermissionError) as e:
+        print(f"kernel_supports_p2pdma() check failed: {e}", file=sys.stderr)
+        return False
 
 
 def hip_runtime_supports_ais():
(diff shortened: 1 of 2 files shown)


tools/ais-check/ais-check
```diff
@@ -12,9 +12,33 @@ If components are missing the program exits with a non-zero exit code.
...
+        if os.path.exists(path):
+            with opener(path, "rt") as f:
+                for line in f:
+                    if line.startswith("CONFIG_PCI_P2PDMA=y") or line.startswith(
+                        "CONFIG_PCI_P2PDMA=m"
+                    ):
+                        return True
+    return False
+
+
```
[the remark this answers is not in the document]
sbates130272: @daforbes-amd this copilot suggest seems more robust to me as it catches both the case where `/proc/config` exists and when the config lives in `/boot/`.

tools/ais-check/ais-check
```diff
@@ -12,9 +12,33 @@ If components are missing the program exits with a non-zero exit code.
...
 
 
+def kernel_supports_p2pdma():
+    """
+    Check for P2P DMA support in the kernel
+    """
+
+    kr = subprocess.check_output(["uname", "-r"]).decode().strip()
+
+    for path, opener in [
```
sbates130272, line 28: @derobins we could also consider checking `/proc/kallsyms` but we would need to ensure the thing we check for is actually not just a stub when `CONFIG_PCI_P2PDMA=n`.

sbates130272 (commented):
@derobins a few comments.

tools/ais-check/ais-check
```diff
@@ -12,9 +12,38 @@ If components are missing the program exits with a non-zero exit code.
...
+    """
+
+    kr = os.uname().release
+
+    for path, opener in [
+        (f"/boot/config-{kr}", open),
+        (f"/lib/modules/{kr}/build/.config", open),
+        ("/proc/config.gz", gzip.open),
+    ]:
+        if os.path.exists(path):
```
jordan-turbofish: You could omit the path exists check, since you'll catch the error when you try to open it anyways.

tools/ais-check/ais-check
```diff
@@ -12,9 +12,38 @@ If components are missing the program exits with a non-zero exit code.
...
+                with opener(path, "rt") as f:
+                    for line in f:
+                        if line.startswith("CONFIG_PCI_P2PDMA=y") or line.startswith(
+                            "CONFIG_PCI_P2PDMA=m"
+                        ):
+                            return True
+            except OSError as e:
+                print(f"kernel_supports_p2pdma() check failed: {e}", file=sys.stderr)
+                continue
+
```
jordan-turbofish, line 43: Might be worth printing a message if we don't find any config files.

tools/ais-check/ais-check
derobins: I thought about that, but I've always been told that it's bad form to use exceptions for control flow.

tools/ais-check/ais-check
derobins: Done

tools/ais-check/ais-check
jordan-turbofish: I think it kind of depends on the context.  This pattern of checking then using can be bad and lead to toctou errors: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

While it shouldn't be an issue in this case, I think its good to try and stick to the secure pattern.

riley-dixon: FWIW, Python prefers the [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) approach when it comes to dealing with exceptions like this.

derobins: This is fixed now

riley-dixon (approved):

jordan-turbofish (approved):

PR #215, opened 2026-03-11
Backend: Add BackendWithFallback to retry IO

riley-dixon:
## Motivation

Modifies Backend & adds a new BackendWithFallback interface to re-attempt an IO request that fails with a different Backend. For example, if a Fastpath IO request fails, the Fallback path can automatically retry the IO.

## Technical Details

* Derived `BackendWithFallback` backends can optionally register another Backend as eligible to retry failed IO requests.
* hipFile will resubmit the failed IO request if a fallback backend is registered and would otherwise accept the original IO request.
* Backend has been reworked for how IO requests are processed. A new virtual function `_io_impl()` has been added which handles only the logic required to submit the IO request to the kernel/driver. `io()` now acts as the entry point and can handle coordinating other tasks such as updating stats or the auto fallback mechanism. (Part of the rationale for this change is so the auto fallback mechanism only retries the IO if the IO failed and not something else.)

## Test Plan

Added unit tests and Fastpath integration level tests. System level tests will be added separately as it is a non-trivial addition.

### Reviewer Note
When reviewing the first couple of commits, you may wish to be aware of certain changes made later on.

AIHIPFILE-100


test/amd_detail/fastpath.cpp
```diff
@@ -555,4 +559,124 @@ TEST_P(FastpathIoParam, IoSizeIsTruncatedToMaxRWCount)
...
+        // Can't use EXPECT_THROW due to the thrown exception type only being known at runtime
+        try {
+            fastpath_backend->io(_get_param_io_type(), mfile, mbuffer, DEFAULT_IO_SIZE, 0, 0);
+        }
+        catch (const std::exception &actual_exc) {
+            // Note: std::exception_ptr preserves the original thrown exception rather than
+            //       creating a new one. This allows us to verify that the expected and
+            //       actual exceptions refer to the same underlying exception object.
+            ASSERT_EQ(&expected_exc, &actual_exc);
+        }
```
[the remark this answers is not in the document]
riley-dixon: This is true. Additionally, since the exception type is parameterized, this makes it an additional challenge to preserve type information when setting the expectation to Throw an exception.

src/amd_detail/backend/fallback.cpp
```diff
@@ -54,12 +54,26 @@ ssize_t
 Fallback::io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer, size_t size,
              hoff_t file_offset, hoff_t buffer_offset)
 {
-    return io(type, file, buffer, size, file_offset, buffer_offset, DefaultChunkSize);
+    return _io_impl(type, file, buffer, size, file_offset, buffer_offset, DefaultChunkSize);
```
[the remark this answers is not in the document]
riley-dixon: Fixed so that Fallback now calls stats.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,84 @@
...
+        case (IoType::Read):
+            update_read_stats(nbytes);
+            break;
+        case (IoType::Write):
+            update_write_stats(nbytes);
+            break;
+        default:
+            break;
+    }
+    return nbytes;
```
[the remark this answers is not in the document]
riley-dixon: Will add the guard to check for nbytes >= 0 , however I agree we may want to look at what we are returning. To my knowledge we only throw exceptions on errors at this moment in time.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,84 @@
...
+bool
+BackendWithFallback::is_fallback_eligible(std::exception_ptr e_ptr, ssize_t nbytes,
+                                          std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer,
+                                          size_t size, hoff_t file_offset,
+                                          hoff_t buffer_offset) const noexcept
+{
+    (void)e_ptr;
+    (void)nbytes;
+    return static_cast<bool>(fallback_backend) &&
+           fallback_backend->score(file, buffer, size, file_offset, buffer_offset) >= 0;
```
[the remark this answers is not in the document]
riley-dixon: Will remove

riley-dixon:
@copilot

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
```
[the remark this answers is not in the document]
riley-dixon: Forgot to remove, will remove.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,88 @@
...
+    catch (...) {
+        std::exception_ptr e_ptr = std::current_exception();
+        if (is_fallback_eligible(e_ptr, nbytes, file, buffer, size, file_offset, buffer_offset)) {
+            nbytes = fallback_backend->io(type, file, buffer, size, file_offset, buffer_offset);
+        }
+        else {
+            throw;
+        }
+        return nbytes;
+    }
```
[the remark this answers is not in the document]
riley-dixon: I am hesitant to apply this as other Backend's may have different criteria for what arguments are permitted.

test/amd_detail/fastpath.cpp
```diff
@@ -555,4 +559,127 @@ TEST_P(FastpathIoParam, IoSizeIsTruncatedToMaxRWCount)
...
+    // Have to rethrow the exception_ptr to be able to access the exception
+    // This looks ugly, but is better than the alternative of trying to preserve the
+    // exception type when setting Throw(*std::shared_ptr<std::exception>>).
+    try {
+        std::rethrow_exception(_get_param_exc_ptr());
+    }
+    catch (const std::exception &expected_exc) {
+        // Can't use EXPECT_THROW due to the thrown exception type only being known at runtime
+        try {
+            fastpath_backend->io(_get_param_io_type(), mfile, mbuffer, DEFAULT_IO_SIZE, 0, 0);
```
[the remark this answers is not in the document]
riley-dixon: Added

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,87 @@
...
```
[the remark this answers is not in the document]
riley-dixon: Similar in nature to one of the previous CoPilot suggestions.

src/amd_detail/backend.cpp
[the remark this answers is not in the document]
riley-dixon: Removed the docstring that a negative integer can be returned. While this is not enforced by the compiler (since `ssize_t` is signed), the Fallback mechanism will check for a negative result anyways. Other nbytes >= 0 checks have been reverted.

src/amd_detail/backend.cpp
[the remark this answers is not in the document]
riley-dixon: We have decided to remove the mention of a negative error code can be returned from the io() docstring.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+    (void)e_ptr;
+    (void)nbytes;
+    return static_cast<bool>(fallback_backend) &&
+           fallback_backend->score(file, buffer, size, file_offset, buffer_offset) >= 0;
+}
+
+void
+BackendWithFallback::register_fallback_backend(std::shared_ptr<Backend> backend) noexcept
+{
```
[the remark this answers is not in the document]
kurtmcmillan: I /mostly/ agree with copilot. register_fallback_backend() should throw an exception if self-registration is attempted

src/amd_detail/backend/fastpath.cpp
```diff
@@ -184,15 +184,17 @@ Fastpath::io(IoType type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer, si
         default:
             throw std::runtime_error("Invalid IoType");
     }
-    switch (type) {
-        case IoType::Read:
-            statsAddFastPathRead(nbytes);
```
kurtmcmillan, line 205: I think it would be better if `statsAddFastPathRead(nbytes)` would be moved after `Context<Hip>::get()->hipAmdFileRead(handle, devptr, size, file_offset);`.

If `hipAmdFileRead` doesn't return a positive value it throws so stats would never be incremented.

I don't think a `update_read_stats` method should be added to `Backend` nor `RetryableBackend`

Same goes for the write stats.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+    catch (...) {
+        std::exception_ptr e_ptr = std::current_exception();
+        if (is_retryable(e_ptr, nbytes)) {
+            nbytes = retry_backend->io(type, file, buffer, size, file_offset, buffer_offset);
+        }
+        else {
+            throw;
+        }
+        // For now, assume retry_backend will handle its own stats (which is true at this moment).
```
kurtmcmillan: Yes each backend should handle its own stats.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return nbytes;
+}
+
+bool
+RetryableBackend::is_retryable(std::exception_ptr e_ptr, ssize_t nbytes) const
+{
+    (void)e_ptr;
+    (void)nbytes;
+    return static_cast<bool>(retry_backend);
+}
```
kurtmcmillan: I think this function should be pure virtual so that implementers of `RetryableBackend` have to implement it.

src/amd_detail/backend.h
```diff
@@ -58,4 +58,54 @@ struct Backend {
...
+    /// @note By default, RetryableBackend just checks if a Backend has been
+    ///       registered for retrying an IO.
+    ///
+    /// @return True if this RetryableBackend can retry the IO, else False.
+    virtual bool is_retryable(std::exception_ptr e_ptr, ssize_t nbytes) const;
+
+    /// @brief Register a Backend to call into to retry a failed IO operation.
+    ///
+    /// @param backend Backend to retry a failed IO operation.
+    void register_retry_backend(std::shared_ptr<Backend> backend) noexcept;
```
kurtmcmillan: This should throw for self-registrations and invalid values (nullptr)

src/amd_detail/state.cpp
```diff
@@ -275,11 +275,18 @@ std::vector<std::shared_ptr<Backend>>
 DriverState::getBackends() const
 {
     static bool once = [&]() {
-        if (Context<Configuration>::get()->fastpath()) {
```
kurtmcmillan: This will have to change with how backends are enabled/disabled now.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,63 @@
...
+RetryableBackend::io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer, size_t size,
+                     hoff_t file_offset, hoff_t buffer_offset)
+{
+    ssize_t nbytes{0};
+    try {
+        nbytes = retryable_io(type, file, buffer, size, file_offset, buffer_offset);
+    }
+    catch (...) {
+        std::exception_ptr e_ptr = std::current_exception();
+        if (is_retryable(e_ptr, nbytes)) {
```
kurtmcmillan: Were you also going have the retry-backend score the IO to see if it was willing to perform the IO?

test/amd_detail/backend.cpp
```diff
@@ -0,0 +1,108 @@
...
+TEST_P(HipFileRetryableBackend, IOSuccess)
+{
+    EXPECT_CALL(*default_backend, retryable_io).WillOnce(Return(successful_io_size));
+
+    ssize_t nbytes = default_backend->io(io_type, mock_file, mock_buffer, 0, 0, 0);
+
+    ASSERT_EQ(nbytes, successful_io_size);
+}
+
+TEST_P(HipFileRetryableBackend, IOFailureNoFallback)
```
kurtmcmillan: Maybe the test name could indicate that there is no retry if there is no registered retry backend

src/amd_detail/backend.cpp
```diff
@@ -26,7 +26,7 @@ RetryableBackend::io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<I
     }
     catch (...) {
         std::exception_ptr e_ptr = std::current_exception();
-        if (is_retryable(e_ptr, nbytes)) {
+        if (is_retryable(e_ptr, nbytes, file, buffer, size, file_offset, buffer_offset)) {
```
kurtmcmillan: `is_retryable` is asking if that specific backend has rejected it for a specific reason. that may not be valid for another backend (i.e. fastpath's device lookup fails).

I don't think we should wrap another backend's scoring into this. This also will not be possible if `is_retryable` is made pure virtual.

I think this should be:
```
if (is_retryable(e_prt, nbytes) && 0 < retry_backend->score(...))
{
   // retry the IO
}
```

test/amd_detail/mbackend.h
```diff
@@ -23,7 +23,6 @@ struct MBackend : Backend {
 struct MRetryableBackend : RetryableBackend {
     MOCK_METHOD(int, score, (std::shared_ptr<IFile>, std::shared_ptr<IBuffer>, size_t, hoff_t, hoff_t),
                 (const, override));
-    MOCK_METHOD(bool, is_retryable, (std::exception_ptr e_ptr, ssize_t nbytes), (const, override));
```
kurtmcmillan: I think you still need to have the mock

src/amd_detail/backend/fallback.cpp
```diff
@@ -54,12 +54,26 @@ ssize_t
...
 }
 
 ssize_t
-Fallback::io(IoType io_type, shared_ptr<IFile> file, shared_ptr<IBuffer> buffer, size_t size,
+Fallback::io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer, size_t size,
```
kurtmcmillan, line 56: You shouldn't need `Fallback::io`. `Backend::io` will call `Fallback::_io_impl (without chunksize)` which will then call `Fallback::_io_impl (with chunksize)`.

src/amd_detail/backend.h
```diff
@@ -55,7 +55,28 @@ struct Backend {
     ///
     /// @throws Hip::RuntimeError Sys::RuntimeError
     virtual ssize_t io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer, size_t size,
-                       hoff_t file_offset, hoff_t buffer_offset) = 0;
+                       hoff_t file_offset, hoff_t buffer_offset)
+    {
```
kurtmcmillan: The implementation for `Backend::io` should go in `backend.cpp`

src/amd_detail/backend/fastpath.cpp
riley-dixon: I wanted to keep updating stats outside of `_io_impl` so that we don't retry the IO if updating the stats fails. Would you be okay with that effect, or alternatively adding a try-catch block wrapping the stats call?

src/amd_detail/backend.cpp
riley-dixon: This was handled in a later commit and the comment deleted.

src/amd_detail/state.cpp
riley-dixon: Updated in a (soon-to-be-pushed) merge commit. If you would like to see a more sophisticated change please let me know.

_Edit: Now pushed._

src/amd_detail/backend.cpp
riley-dixon: Yes, this is done in a subsequent commit when `is_retryable` was reworked.

test/amd_detail/fastpath.cpp
```diff
@@ -555,4 +559,73 @@ TEST_P(FastpathIoParam, IoSizeIsTruncatedToMaxRWCount)
...
+        return std::get<0>(GetParam());
+    }
+
+    inline const std::exception_ptr _get_param_exc_ptr() const
+    {
+        return std::get<1>(GetParam());
+    }
+};
+
+// The Fastpath can throw a few different kinds of derived std::runtime_errors.
```
kurtmcmillan: I think this test needs to be broken up so that each test focuses on specific functionality.

I think what you could do is instantiate a fastpath backend instance and then register a mock backend as the fallback backend. then you could have a number of tests where each test ensures that `Fastpath::is_retryable` does the right thing for different return values from  `Hip::hipAmdFileRead`/`Hip::hipAmdFileWrite` (`-ENODEVICE`, ....).

src/amd_detail/backend.cpp
riley-dixon: I see the benefit of calling `score()` here instead.

Though I am confused about your comment about "this also will not be possible..." since the derived class would have access to the fallback backend reference.

src/amd_detail/backend.cpp
riley-dixon: Makes sense in conjunction with https://github.com/ROCm/hipFile/pull/215#discussion_r2984151338

src/amd_detail/backend/fastpath.cpp
kurtmcmillan: Updating stats shouldn't fail/throw. Those functions are just bumping a counter. stats should be fixed.

src/amd_detail/backend/fallback.cpp
riley-dixon: This change was made in conjunction with changing how stats are called in Fallback. I apologize for the confusion since this commit is missing that change.

src/amd_detail/backend.cpp
kurtmcmillan: Yeah that isn't very clear and even this morning I'm having trouble understanding what I meant.

I think that i mean that having `is_retryable` calling the retry backend's score function will not be possible if `is_retryable` is pure virtual and implemented by Fastpath. Fastpath will not have access to the retry backend's score function so calling the retry backend's score within `is_retryable` will not be possible.

I hope that is more clear.

src/amd_detail/backend/fallback.cpp
kurtmcmillan: Looking at the current state of changes, the stats update should be at the end of `Fallback::_io_impl` (the one with chunksize). If the function is successful the stats are bumped, if the IO fails the stats don't get bumped.

src/amd_detail/backend/fastpath.cpp
zbyrne: i'll update all the stats calls to noexcept.

src/amd_detail/backend.cpp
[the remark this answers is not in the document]
riley-dixon: Added a check for self-registration & nullptr. I don't think we need something more comprehensive right now.

src/amd_detail/backend/fastpath.cpp
riley-dixon: Stats changes have been reverted with the assumption that stats won't throw.

src/amd_detail/backend.cpp
riley-dixon: Made pure virtual.

src/amd_detail/backend.h
riley-dixon: Added these checks.

test/amd_detail/backend.cpp
riley-dixon: I changed the test name, but please let me know if you would like a further change.

src/amd_detail/backend.cpp
riley-dixon: Had a discussion and cleared up the confusion. The suggested change above has been made. I did add a nullptr check on the fallback_backend prior to the `is_fallback_eligible` check.

test/amd_detail/mbackend.h
riley-dixon: Since reworking `is_fallback_eligible` this has been re-added.

src/amd_detail/backend/fallback.cpp
riley-dixon: Yup, reverted the stats change and removed the Fallback:io() override to use the default Backend::io() implementation.

src/amd_detail/backend.h
riley-dixon: Done

test/amd_detail/fastpath.cpp
riley-dixon: I can definitely add some new unit tests specifically for `Fallback::is_fallback_eligible`. I think though I would still like to keep the integration tests if possible. Yes, it touches on a lot of different areas but that is the point, to make sure the fallback mechanism behaves as intended starting from io submission. The future system tests will expand on this by not mocking out all Hip & Sys calls.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+void
+BackendWithFallback::register_fallback_backend(std::shared_ptr<Backend> backend)
+{
+    if(!backend) {
+        throw std::invalid_argument("Cannot register a nullptr as a fallback backend.");
+    } else if (backend.get() == this) {
+        throw std::invalid_argument("Cannot register a backend as it's own fallback.");
+    }
```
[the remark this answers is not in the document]
riley-dixon: Added

test/amd_detail/fastpath.cpp
```diff
@@ -599,4 +603,134 @@ TEST_P(FastpathIoParam, IoSizeIsTruncatedToMaxRWCount)
...
+        try {
+            fastpath_backend->io(_get_param_io_type(), mfile, mbuffer, DEFAULT_IO_SIZE, 0, 0);
+            FAIL() << "io() was expected to throw, but it returned normally";
+        }
+        catch (const std::exception &actual_exc) {
+            // Verify that the propagated exception has the same dynamic type and message
+            // as the one stored in the original std::exception_ptr, without relying on
+            // pointer identity of the underlying exception object.
+            ASSERT_EQ(typeid(expected_exc), typeid(actual_exc));
+            ASSERT_STREQ(expected_exc.what(), actual_exc.what());
```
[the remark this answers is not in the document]
riley-dixon: This should not be an issue because the information in the exception should not change.

test/amd_detail/fastpath.cpp
```diff
@@ -599,4 +603,134 @@ TEST_P(FastpathIoParam, IoSizeIsTruncatedToMaxRWCount)
...
+            EXPECT_CALL(msys, pread).WillRepeatedly(ReturnArg<2>());
+            break;
+        case IoType::Write:
+            // Called by Fastpath
+            EXPECT_CALL(mhip, hipAmdFileWrite).WillOnce(Rethrow(_get_param_exc_ptr()));
+            // Called by Fallback
+            EXPECT_CALL(mhip, hipStreamSynchronize).WillRepeatedly(Return());
+            EXPECT_CALL(msys, fdatasync).WillRepeatedly(Return());
+            EXPECT_CALL(msys, pwrite).WillRepeatedly(ReturnArg<2>());
+            break;
```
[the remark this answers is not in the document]
riley-dixon: No.

src/amd_detail/backend.cpp
```diff
@@ -0,0 +1,62 @@
...
+                        size_t size, hoff_t file_offset, hoff_t buffer_offset)
+{
+    ssize_t nbytes{0};
+    try {
+        nbytes = _io_impl(type, file, buffer, size, file_offset, buffer_offset);
+        if (nbytes < 0) {
+            // Typically we should not reach this point. But in case we do, throw
+            // an exception to use the fallback backend.
+            throw std::system_error(-static_cast<int>(nbytes), std::generic_category());
+        }
```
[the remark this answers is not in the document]
riley-dixon: nbytes in this context is errno. Since we don't want to change our return type, this is the approach we decided on.

src/amd_detail/backend/fallback.cpp
kurtmcmillan: thanks.

test/amd_detail/fastpath.cpp
kurtmcmillan: This test introduces implementation details of fallback into fastpath tests. Instead of registering the fallback backend as the fallback, could you register a mock backend as the fallback?

There should also be unit tests to ensure that `is_fallback_eligible` returns as expected for different inputs.

riley-dixon: `is_fallback_eligible` is not directly accessible since we discussed moving its visibility to be protected. Other unit tests have already been created to cover different scenarios.

> could you register a mock backend as the fallback?

There are already unit tests that do this. The integration test is specific to test that the Fastpath->Fallback IO path works as intended with as much real code as possible. At this point I'll remove the integration test, but we should get the system test in so that we are covering this execution path.

src/amd_detail/backend.h
```diff
@@ -58,11 +59,55 @@ struct Backend {
...
+// different Backend in the event of an error.
+struct BackendWithFallback : public Backend {
+    ssize_t io(IoType type, std::shared_ptr<IFile> file, std::shared_ptr<IBuffer> buffer, size_t size,
+               hoff_t file_offset, hoff_t buffer_offset) override final;
+
+    /// @brief Register a Backend to retry a failed IO operation.
+    ///
+    /// @param backend Backend to retry a failed IO operation.
+    ///
+    /// @throws std::invalid_argument If a nullptr or self references is passed.
```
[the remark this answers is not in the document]
riley-dixon: done

jordan-turbofish (approved):

derobins (approved):

PR #216, opened 2026-03-13
ais-check: Add better HIP detection

derobins:
Adds better HIP detection and output

AIHIPFILE-156

### DIFF
--- a/tools/ais-check/ais-check
+++ b/tools/ais-check/ais-check
@@ -12,9 +12,13 @@ If components are missing the program exits with a non-zero exit code.
 import argparse
 import ctypes
 import ctypes.util
+import glob
 import gzip
 import os
 import sys
+
+# Global list of found HIP runtime libraries
+hip_libraries = {}
 
 
 def kernel_supports_p2pdma():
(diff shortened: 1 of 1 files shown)


tools/ais-check/ais-check
```diff
@@ -46,67 +50,113 @@ def kernel_supports_p2pdma():
...
+                        f"hipGetProcAddress({symbol}) failed with err code"
+                        f" {err} ({err_str}) and symbolStatus"
+                        f" {symbol_status.value}",
+                        file=sys.stderr,
+                    )
+                continue
+
+        hip_libraries[hip_path] = True
+
+    return bool(True in hip_libraries.values())
```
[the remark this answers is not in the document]
derobins: D'oh!

tools/ais-check/ais-check
```diff
@@ -46,67 +50,134 @@
...
+
+    # 4. Versioned installs (/opt/rocm-5.x, etc.)
+    for d in glob.glob("/opt/rocm*/lib*/libamdhip64.so"):
+        candidates.append(d)
+
+    # Drop any paths that don't exist
+    existing_paths = [
+        path
+        for path in candidates
+        if os.path.exists(os.path.abspath(os.path.normpath(path)))
```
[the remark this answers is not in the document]
derobins: I'm not going to restrict the script to a weird subset of directories. That will baffle anyone who installed ROCm into an unusual directory.

kurtmcmillan:
I think that ais-check should return success if the current environment supports the fastpath. I think this change breaks that. If the current environment is using ROCm 7.0, but ROCm 7.2 is also installed, would this script return success? If so, I don't think it should.

tools/ais-check/ais-check
```diff
@@ -46,67 +50,134 @@ def kernel_supports_p2pdma():
...
-            ctypes.byref(func_ptr),
-            version.value,
-            0,
-            ctypes.byref(symbol_status),
-        )
+        hip.hipGetErrorString.argtypes = [hipError_t]
+        hip.hipGetErrorString.restype = ctypes.c_char_p
+
+        version = ctypes.c_int()
+        err = hip.hipRuntimeGetVersion(ctypes.byref(version))
```
riley-dixon, line 142: It's easier to just read `/opt/<rocm>/.info/version` to get the version of the library. This file is provided by the `rocm-core` package which is mandatory.

tools/ais-check/ais-check
```diff
@@ -154,8 +225,17 @@ def main():
         u = os.uname()
         print()
         print(u.sysname, u.nodename, u.release, u.version, u.machine)
+
         print()
+        print("Found these HIP libraries (some may be symlinks):")
```
riley-dixon: Might be worth adding the word "redundant" (e.g. `redundant symlinks`) since /opt/rocm will likely be a duplicate entry (tbh I can't think of typical scenarios where it's not). To fix entirely, we could track ROCm installs by the realpath to get rid of symlinks.

tools/ais-check/ais-check
jordan-turbofish: I don't think that will give you the same information.  For example 6.4.4 gives you `60443484` when called through hipRuntimeGetVersion (6 for major, 04 for minor, and 43484 for patch version).  The version file has `6.4.4-129`.

riley-dixon: Hmm I see - fair enough then, I'll drop this suggestion if there is value to be found in the patch version.

jordan-turbofish:
Played around with getting the absolute path when using find_library:
```
def dlinfo_path(library_name):
    class LinkMap(ctypes.Structure):
        _fields_ = [
            ("l_addr", ctypes.c_void_p),
            ("l_name", ctypes.c_char_p),
        ]

    dl_path = ctypes.util.find_library("dl")
    library_path = ctypes.util.find_library(library_name)
    if dl_path is None or library_path is None:
        return None

    try:
        libdl = ctypes.CDLL(dl_path)
        libwanted = ctypes.CDLL(library_path)
    except OSError:
        return None

    dlinfo = libdl.dlinfo
    dlinfo.argtypes = ctypes.c_void_p, ctypes.c_int, ctypes.c_void_p
    dlinfo.restype = ctypes.c_int

    lmptr = ctypes.POINTER(LinkMap)()
    err = dlinfo(libwanted._handle, 2, ctypes.byref(lmptr))
    if err != 0:
        return None

    abspath = lmptr.contents.l_name.decode('ascii')
    return abspath

print(f"amdhip64 path:{dlinfo_path(library_name='amdhip64')}")
```
Testing:
```
> python3 test.py
amdhip64 path:None
> LD_LIBRARY_PATH=/home/jpatters/hip-7.2.0-dbg/lib:/opt/rocm/lib python3 test.py
amdhip64 path:/home/jpatters/hip-7.2.0-dbg/lib/libamdhip64.so.7
```

ammallya:
Imported to ROCm/rocm-systems pull 4862

PR #228, opened 2026-03-19
Fix unused variables in async test

derobins:
Only used in assert statements, so unused in release builds.

Suppressed using C++17's [[maybe_unused]].


derobins:
Dropped in favor of #229

PR #234, opened 2026-03-23
Rildixon/io retry compat system test

riley-dixon:
GTest's framework allows for a mock to call the real implementation of a method. We can utilize this in our system tests to very selectively choose what gets mocked. In this case, we want to mock the call from KFD/HSA to induce a failure to test the fallback mechanism.

Relies on #215 

AIHIPFILE-155


ammallya:
Failed to import to ROCm/rocm-systems due to merge conflicts please reopen the PR on rocm-systems

PR #242, opened 2026-03-30
Fix high-severity Coverity issues

derobins:
Fixes a couple of high-severity issues identified by Coverity Scan

* Marks use-after-move as intentional in the file descriptor tests
* Fully initialize a struct in the stats code
* Ensure a socket is cleaned up on errors in `StatsServer::threadFn()`
* Check the return value from poll(2) in stats.cpp

Part of AIHIPFILE-161

### DIFF
--- a/src/amd_detail/stats.cpp
+++ b/src/amd_detail/stats.cpp
@@ -146,7 +146,23 @@ StatsServer::threadFn()
     }
     while (true) {
         pollfd pfd[2]{{sock, POLLIN, 0}, {m_efd.get(), POLLIN, 0}};
-        poll(&pfd[0], 2, -1);
+
+        int ret = poll(&pfd[0], 2, -1);
+
+        if (ret == 0) {
+            continue;
+        }
+        else if (ret < 0) {
+            if (errno == EINTR) {
+                // Signal interrupt
+                continue;
+            }
+            else {
+                // Badness
+                break;
+            }
+        }
+
         if (pfd[0].revents & POLLIN) {
             socklen_t addrlen{sizeof(addr)};
             int       conn{accept(sock, reinterpret_cast<sockaddr *>(&addr), &addrlen)};
--- a/src/amd_detail/stats.cpp
+++ b/src/amd_detail/stats.cpp
@@ -134,15 +134,15 @@ StatsServer::threadFn()
     int   sock{socket(AF_UNIX, SOCK_STREAM, 0)};
     pid_t pid{getpid()};
     if (sock == -1) {
-        return;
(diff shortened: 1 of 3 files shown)


src/amd_detail/stats.cpp
```diff
@@ -132,15 +134,15 @@ StatsServer::threadFn()
     int   sock{socket(AF_UNIX, SOCK_STREAM, 0)};
     pid_t pid{getpid()};
     if (sock == -1) {
-        return;
+        goto out;
```
zbyrne: Could you put sock in a FileDescriptor instead of using goto?

derobins: Oh, I see - as in Copilot's suggestion below. I'll implement after lunch.

src/amd_detail/stats.cpp
```diff
@@ -22,8 +22,8 @@ sendFd(int sock, int fd) noexcept
 {
     int      data{1};
     iovec    iov{&data, sizeof(data)};
-    msghdr   msgh;
-    cmsghdr *cmsgp;
+    msghdr   msgh{};
+    cmsghdr *cmsgp = nullptr;
```
zbyrne: {} initialization for consistency

src/amd_detail/stats.cpp
```diff
@@ -54,8 +55,8 @@ recvFd(int sockfd) noexcept
 {
     int      data, fd;
     iovec    iov{&data, sizeof(data)};
-    msghdr   msgh;
-    cmsghdr *cmsgp;
+    msghdr   msgh{};
+    cmsghdr *cmsgp = nullptr;
```
zbyrne: {} here too

zbyrne (approved):

src/amd_detail/stats.cpp
kurtmcmillan: I think it would be good to zero init `controlMsg` as well while we're in the area.

test/amd_detail/stats.cpp
```diff
@@ -55,7 +55,7 @@ TEST_F(HipFileStats, StatsServerLifetime)
     EXPECT_CALL(msys, ftruncate);
     EXPECT_CALL(msys, mmap).WillOnce(testing::Return(&buff));
     EXPECT_CALL(msys, munmap);
-    EXPECT_CALL(msys, close).Times(2);
+    EXPECT_CALL(msys, close).Times(3);
```
derobins, line 58: Have to bump this because now we are calling one more close() through the mock

zbyrne (approved):

PR #243, opened 2026-03-31
Coverity: Set umask 0077 before creating tmp files

derobins:
Coverity points out that it's a potential security problem if you create tmp files with group and world permissions.

Adds umask calls to state_mt and the common test header.

Part of AIHIPFILE-161

### DIFF
--- a/test/amd_detail/state_mt.cpp
+++ b/test/amd_detail/state_mt.cpp
@@ -26,6 +26,7 @@
 #include <unistd.h>
 #include <utility>
 #include <sys/resource.h>
+#include <sys/stat.h>
 #include <vector>
 
 using namespace hipFile;
--- a/test/amd_detail/state_mt.cpp
+++ b/test/amd_detail/state_mt.cpp
@@ -49,6 +49,10 @@ static atomic<bool> run_flag{true};
 static int
 make_temp_file()
 {
+    // Security analyzers will be sad if you don't set umask 0077
+    // before creating temporary files
+    mode_t old_umask = umask(S_IRWXG | S_IRWXO);
+
     // Create a unique temporary file
     char pathname[]{"state_mt_tmp.XXXXXX"};
     int  fd{mkstemp(pathname)};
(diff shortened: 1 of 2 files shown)


test/common/test-common.h
```diff
@@ -62,11 +63,17 @@ struct Tmpfile {
...
+        // before creating temporary files
+        mode_t old_umask = umask(S_IRWXG | S_IRWXO);
+
         path += "/hipFile.XXXXXX";
         if ((fd = mkstemp(path.data())) == -1) {
             throw std::runtime_error("Could not create temporary file");
         }
 
+        umask(old_umask);
+
```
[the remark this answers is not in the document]
derobins: Or we could just call umask before we throw. 1 line of simple, understand able code >> 5 lines of pedantic C++ gibberish. This simple function will probably not be dramatically expanded in the future.

test/amd_detail/state_mt.cpp
```diff
@@ -369,6 +370,15 @@ thread_function(int id)
...
 main()
 {
+    // Security analyzers will be sad if you don't set umask 0077
+    // before creating temporary files
+    //
+    // Doing this here to avoid inadvertently synchronizing files if we
+    // added a mutex in the temp file function
+    //
+    // Ignoring return type since we never switch it back
+    (void)umask(S_IRWXG | S_IRWXO);
```
riley-dixon, line 380: Is there a reason why the security analyzer is recommending 0077 over 0066 for tmpfiles? I don't think we are creating executables here?

riley-dixon (approved):

test/amd_detail/state_mt.cpp
derobins: I think 0077 is the standard security-recommended permission level for temporary files

derobins: In the grand scheme of things, this is overkill. This is just test code - we aren't putting sensitive data in /tmp.

riley-dixon: Oh for sure. Just seemed a bit odd to me but not going to question it to hard.

PR #247, opened 2026-04-02
Coverity: Fix a few uninitialized variables

derobins:
Part of AIHIPFILE-161

### DIFF
--- a/test/system/async.cpp
+++ b/test/system/async.cpp
@@ -243,9 +243,9 @@ public:
         ASSERT_EQ(hipStreamDestroy(hip_stream), hipSuccess);
         ASSERT_EQ(hipFree(dev_ptr), hipSuccess);
     }
-    void                            *dev_ptr;
+    void                            *dev_ptr{};
     Tmpfile                          tf;
-    hipFileHandle_t                  fh;
+    hipFileHandle_t                  fh{};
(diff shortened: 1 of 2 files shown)


test/system/async.cpp
```diff
@@ -243,24 +243,24 @@ class HipAsyncMemcpyKernel : public ::testing::Test {
...
     std::shared_ptr<IFile>           file;
     std::shared_ptr<IBuffer>         buffer;
     std::shared_ptr<IStream>         stream;
     std::shared_ptr<AsyncOpFallback> op;
-    void                            *op_dev_ptr;
-    void                            *kernel_args[1];
-    int                              max_threads_per_block;
+    void                            *op_dev_ptr{};
+    void                            *kernel_args[1]{};
+    int                              max_threads_per_block = 0;
```
kurtmcmillan, line 263: in the future this could be:
```
int                              max_threads_per_block{0};
```
or
```
int                              max_threads_per_block{};
```

kurtmcmillan (approved):

test/system/async.cpp
derobins: I thought about that, but went with consistency. We can do a global fix of things like that in a separate PR.

